Issue #7, opened 2019-12-10
Include permalink when using the Post-action todo action


aaronrothschild:
Tasks made from the Post "..." menu don’t include a link to the originating convo for maintaining context about the todo item (it would help if you added/concatenated a permalink to the list item when it is created via post-action command).

larkox:
May I take this one?

hanzei:
Sure, thanks @larkox

PR #10, opened 2019-12-13
(wip) Go 1.13 and fixing the tests

levb:

PR #17, opened 2019-12-24
[GH-7] Include permalink when using the Post-action todo action

larkox:
#### Summary

Include permalink when using the Post-action todo action
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-todo/issues/7


jasonblais:
@larkox Thanks for the contribution! I queued the PR for reviews, but a heads up that they may take longer than usual with the holidays.

You probably already noticed, but looks like the tests failed for these changes.

webapp/src/selectors.js
```diff
@@ -17,3 +19,34 @@ export const getMessage = (state) => {
...
+export const getCurrentTeamRoute = (state) => {
+    const config = getConfig(state);
+
+    const basePath = getSiteURL();
+    const teamName = state.entities.teams.teams[state.entities.teams.currentTeamId].name;
+
+    return basePath + '/' + teamName + '/';
+}
+
+function getSiteURL() {
```
levb: I am a webapp kook, but isn't there a way to get it from a mattermost-provided API?

webapp/src/components/root/root.jsx
```diff
@@ -27,6 +27,7 @@ export default class Root extends React.Component {
     static getDerivedStateFromProps(props, state) {
         if (props.visible && state.message == null) {
             return {message: props.message};
+            // return {message: props.message + ' (' + createPermalink(props.postID) + ')'};
```
levb: nit: remove the comment?

levb:
> You probably already noticed, but looks like the tests failed for these changes.

Yes, unrelated, @jwilander is to fix https://github.com/mattermost/mattermost-plugin-todo/pull/10

webapp/src/selectors.js
larkox: I am not sure how to do this, since the function I copied is from mattermost-webapp, and I am not sure whether it can be imported or how (I tried but I was not able).

If you can direct on how to do this, I am more than happy to make the change.

levb:
I see 
image now - 1/5 can you please make it just "<a...>permalink</a>" for the permalink?

larkox:
Improved. Current version still have one problem.
If we were to use markdown to a permalink ```[link](http://something.com/team/pl/id)``` this permalink will also be parsed. I could use different set of braces or limits for the link instead of parenthesis, or could only parse the last link, but would be in detriment of the possibilities of re-writing the todo the way we want.

What do you think @levb?

aaronrothschild:
>  I could use different set of braces or limits for the link instead of parenthesis, or could only parse the last link, but would be in detriment of the possibilities of re-writing the todo the way we want.

@larkox Sorry, could you expand on this?  I'm not following?  Thanks!

larkox:
> @larkox Sorry, could you expand on this? I'm not following? Thanks!

Sorry for the chaotic explanation.

In one hand, instead of using normal parenthesis to write the link, we could use any other element to write the link (curly braces, quotes, or anything you may consider). That way it would not have any conflict with markdowns like ```[name](link)```.

Regarding parsing only the last link, it might be not optimal if we want to edit the TODO. For example...
User1 writes:
```[This](link to video) is an awesome video, you should check it.```
I make a todo out of that message, and I have on the textbox the following:
```[This](link to video) is an awesome video, you should check it.
(link to message)```
Since the message just like that is not useful for me, I want to change it to something like this:
```User1 said here (link to message) to watch this (link to video)```
If we were to only parse the last link, we would only parse the link to video, and not the link to the message.

**Comment by levb** [comment]
Also, the build errors should be relevant now

**Comment by levb** [comment]
@aaronrothschild @jfrerich @DHaussermann ping - we may be cutting a release and deploying it soon, would be good to get this improvement in?

**Comment by jfrerich** [comment]
@larkox if you wouldn't mind, can you address the following issue:
image

File: webapp/src/selectors.js [hunk 3]
```diff
@@ -17,3 +17,32 @@ export const getMessage = (state) => {
...
 export const getItems = (state) => getPluginState(state).items;
+export const getCurrentTeamRoute = (state) => {
+    const basePath = getSiteURL();
+    const teamName = state.entities.teams.teams[state.entities.teams.currentTeamId].name;
+
+    return basePath + '/' + teamName + '/';
+};
+
+function getSiteURL() {
+    let siteURL = '';
```
jfrerich: 1/5 suggestion to set the value first and remove the `else`
```suggestion
    let siteURL = siteURL = window.location.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : '');
```
or `siteURL = window.location.origin` and use `if (!window.location.origin)`

jfrerich:
There is one other issue I see, but if it seems difficult to add, we can create it as a separate issue!

image

larkox:
I "solved" the todo list by using quotes instead of bullet points, but I am not sure that is the best approach. Happy to change it to something else if you have a better idea.

The solution formatting is still a bit off sometimes, specially with line breaks, but since it is not the scope of this PR, maybe we should revise it on a different ticket.

jfrerich:
@larkox I like the previous fix you had implemented with the bulleted list, instead of quotes.  Can you revert back to that change and we'll file a separate ticket to handle the misalignment added from the permalink and markdown issues?  Thanks!

larkox:
@jfrerich No problem. Just reverted.

levb:
@aaronrothschild @DHaussermann ready for your approval.

jfrerich:
@DHaussermann gentle reminder this is ready for review

DHaussermann:
Hi @larkox, I see this behaving as expected. Saving a post to the To Do list now created a permalink as expected.

That being said, using permalinks that where added to the list does not quite work I would expect.    This may not be an issue specific to this plugin but I'm seeing that when I click the link from a browser, it opens in a new tab rather then just redirecting the same tab as it would if you had clicked a permalink from a reply thread or flagged posts etc..   

From the desktop app (4.4.0 on Windows) clicking permalinks from the to do list does not seem to do anything.

Any thoughts on how this issue could be solved? Does this require a more complex solution beyond the scope of this PR?

larkox:
I have checked it, and I can "easily" make it load on the same page, instead of a new one. Regarding not loading the page again and just go to the right part of the page, [I am asking the people from Toolkit](https://community.mattermost.com/core/pl/xau4sggk6inst8ifnm9jgd3z8r) to see if there is a straightforward way to do it.
I am checking workarounds, but I have not yet found the way to make it work.

larkox:
Finally discovered the problem, and this would not be possible unless we make changes on Webapp, since we need to access the history object they are using. More information in the previos conversation.

The change to be made in Webapp is almost trivial (export the variable along with the rest of the exports) but I suppose making that design decision may take time, and requiring certain webapp version to run the plugin may not be as desirable.

That being said, if we were to go with this PR without the changes in Webapp, I think that the experience feel better when the link opens in a new site instead of a refreshing the current site. What do you think?

levb:
@larkox @aaronrothschild 0/5 I'd say let's make the change (submit a PR) to the webapp and start the process. Let's see what the webapp team will have to say about the change, and then decide if we really need short-term mitigation?

DHaussermann:
@larkox clicking on permalinks is working now after the webapp change but, in the Windows Desktop app I see a related JS error related to `browserhistory`. Note that it makes the  bar appear in the UI. 

ToDo-JS

This does not occur for me in Chrome.
Please advise if you can repo this.

DHaussermann:
@larkox after re-installing both v4.3.2 and 4.4-rc4 I can no longer see this issue on my Mac.
This change in now working as expected. I will create a WebApp bug if I see this occur again.

hanzei:
Your are awesome @larkox :man_dancing:

Issue #19, opened 2020-01-07
Todo text should support Markdown


levb:

larkox:
I will take this one

PR #20, opened 2020-01-10
[GH-19] Add markdown support to Todos

larkox:
#### Summary

Add markdown support to Todos
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-todo/issues/19


larkox:
In order to add markdown, I have taken a few utils files from webapp, and stripped them down from many things that were not needed for this, and were using other webapp components.

Due to this stripping, some potential bugs may have been created, but should not be the case.

larkox:
Made changes to use webapp exported PostUtils instead of creating our own markdown parser in my latest commit. Now it is simpler, cleaner, and should work better (since it is using webapp version)

levb:
@DHaussermann I already tested this change minimally (built the plugin and tried some markdown) and it appears to work.

Issue #8, opened 2019-12-10
Idea: `/todo` returns the todo list instead of help text


aaronrothschild:
I feel like /todo should just return the list of todos, not show help.  Thats' what `/todo help` should be for

Bonus: It would be really cool if the todo list was displayed in the RHS when you typed /todo it would give the user a place to keep the list in view and go about their work in MM and check off the tasks as they get done

PR #26, opened 2020-01-19
command - /todo  returns the list of todos

thapakazi:
#### Summary
This pull request sets `/todo` to return the list of todos as default, it used to return a help text earlier. 

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-plugin-todo/issues/8
issue_8_todo

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -62,7 +62,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 	switch command {
 	case "add":
 		handler = p.runAddCommand
-	case "list":
+	case "list", "":
 		handler = p.runListCommand
 	case "pop":
 		handler = p.runPopCommand
--- a/server/command.go
+++ b/server/command.go
@@ -49,31 +49,30 @@ func getCommandResponse(responseType, text string) *model.CommandResponse {
 
 // ExecuteCommand executes a given command and returns a command response.
 func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
-	stringArgs := strings.Split(args.Command, " ")
-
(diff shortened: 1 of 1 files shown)


server/command.go
```diff
@@ -62,7 +62,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 	switch command {
 	case "add":
 		handler = p.runAddCommand
-	case "list":
+	case "list", "":
```
levb: Fascinating...

1) @iomodo I was surprised to see this actually work, apparently args.Command comes in with a space appended at the end. We can argue that the plugin should be `strings.TrimSpace`-ing the command, but that's besides the point. This is probably a bug/undesired feature in the server code somewhere. normalizing the command?
2) @thapakazi I recommend not relying on the trailing " ", and having an empty stringArgs[1]; if anything that's a bug; the right place to insert `p.runListCommand` is line 55; I also recommend adding a `strings.TrimSpace` around `args.Command` in line 52.

iomodo: @levb there are 2 problems:
1. Current implementation of the autocomplete appends trailing space automatically when you type enter after a trigger word. So if you want to type just `/todo` you have to delete the space.
2. Server code assumes there will be a trailing space after the trigger word. There's already a [PR](https://github.com/mattermost/mattermost-server/pull/13646) changing that behavior.

levb:
@jfrerich @DHaussermann ping

levb:
@iomodo I am not sure what's the overall direction: can this code be committed, or will it break some day because of the extra space treatment?

@thapakazi perhaps it's worth expressing the logic differently (strings.TrimSpace, and then checking for len(split) == 1?)

thapakazi:
Yesh, I see what you are saying Lev, 
Ieme do the changes and ensure it won't break n the future.

Issue #12, opened 2019-12-13
Request a task from another user


levb:
Sender:
- `/todo request @user some-task`
- the task is added to the sender's todo list in the new "Sent" section; its status is displayed (pending/active). Maybe: display the position (number) in the recipient's "Requests Received" list -- this can be labelled "Priority": number, or "Received", "Declined", "Enqueued"
- When the recipient acts on the task, the sender is notified with a DM from the bot, and the task is updated (moved to "Completed") or removed from the list

Recipient:
- receives a DM from the Bot, notifying her of the new task, with action buttons.
- The task is displayed in the recipient's "Inbox" section, with applicable actions, "╳ (decline)", "🐢 (enqueue)", "✓ Done"

levb:
cc @aaronrothschild for comments

aaronrothschild:
> cc @aaronrothschild for comments

I love the suggestion!  Nothing comes to mind right now, but will think about this topic.

larkox:
May I take this one?

cpoile:
@larkox Great, go to it! :)

larkox:
Here is a mock-up for the front-end. It would be nice to confirm if this is the idea we have in mind before starting to program it down. Keep in mind that this is a mock-up, and most of the shapes and colours are just placeholders. Feel free to propose improvements on that part.

Since there are two new sections (inbox and sent) I named the natural one "own", as in "my own todos".

This section will change slightly. Now, instead of just a button to delete the todos, we will have more accessible buttons for "Done" and "Delete". This is because some Todos will be Todos sent by other users, and we have to make the distinction whether we have done them or we just delete them.

Also, the "Created" changes to "Received from" in case the Todo is received from other user.

![image](https://user-images.githubusercontent.com/1933730/73436528-f842f500-434a-11ea-98db-a55ba7999b7d.png)


On inbox, so far I have used buttons with text. Should we use both text and images ("✓ Done" vs. "Done") only text, or only images? I would say both is better. Do we have any image we want to use? We could try to use font awesome icons like:
- Check for accept: https://fontawesome.com/icons/check?style=solid
- Cross for decline: https://fontawesome.com/icons/times?style=solid
- Plus for enqueue: https://fontawesome.com/icons/plus?style=solid

I also found nice to put the sender along the timestamp, changing "Created" by "Received".

Finally we would remove the X from the items, since we want them to decline the item, not just remove them.
[...]

Issue #4, opened 2019-12-06
Use Bot Account to post slash command responses


JtheBAB:
Hi @jwilander 

When i add a todo or list my todo's it doesn't show the bot logo:

![Clipboard02](https://user-images.githubusercontent.com/42135332/70319807-7be8d580-1823-11ea-9491-2bc31664801b.jpg)

thapakazi:
Is this fixed? 
I could not reproduce this on 
 - localhost (old version: 5.17.0-rc1)
 and 
- [community mattermost](https://community.mattermost.com) (5.18.0)

levb:
@JtheBAB can you please re-test?

JtheBAB:
Daily Reminder is ok:
![image](https://user-images.githubusercontent.com/42135332/72713295-15b6d880-3b6d-11ea-8e7e-15110ca1ebf4.png)

/todo list not:
![image](https://user-images.githubusercontent.com/42135332/72713325-249d8b00-3b6d-11ea-8fa5-176092c42218.png)

same also on community

hanzei:
The code :tm: states that there has never been an icon for the slash command responses: https://github.com/mattermost/mattermost-plugin-todo/blob/master/server/command.go#L46

The fix should be to us the bot account to create the slash command Responses.

attilamolnar:
Hey, I'll take a look

hanzei:
Awesome, thanks @attilamolnar :+1:

PR #27, opened 2020-01-30
[GH-12] Request a task from another user

larkox:
#### Summary
Request a task from another user

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-todo/issues/12


server/command.go
```diff
@@ -98,6 +155,8 @@ func (p *Plugin) runAddCommand(args []string, extra *model.CommandArgs) (*model.
 		Message:  message,
 	}
 
+	p.itemLock.Lock()
```
levb: ditto

server/command.go
```diff
@@ -120,7 +179,19 @@ func (p *Plugin) runAddCommand(args []string, extra *model.CommandArgs) (*model.
 }
 
 func (p *Plugin) runListCommand(args []string, extra *model.CommandArgs) (*model.CommandResponse, bool, error) {
-	items, err := p.getItemsForUser(extra.UserId)
+	listID := OwnListKey
```
levb: 👍

server/command.go
```diff
@@ -120,7 +179,19 @@ func (p *Plugin) runAddCommand(args []string, extra *model.CommandArgs) (*model.
...
+	if len(args) > 0 {
+		switch args[0] {
+		case "inbox":
+			listID = InboxListKey
+		case "sent":
+			listID = SentListKey
+		}
```
levb: default: error/help?

server/command.go
```diff
@@ -120,7 +179,19 @@ func (p *Plugin) runAddCommand(args []string, extra *model.CommandArgs) (*model.
...
+	}
+
+	p.itemLock.RLock()
```
levb: why? :)

server/plugin.go
```diff
@@ -91,14 +99,29 @@ func (p *Plugin) handleAdd(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+type listAPIRequest struct {
+	list string
```
levb: 1/5 `name` or `ID`? Would beed to be Public to `json`

levb: Also, is it even used? Looks like you're using a query arg.

server/plugin.go
```diff
@@ -138,6 +161,157 @@ func (p *Plugin) handleList(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "Not authorized", http.StatusUnauthorized)
+		return
+	}
+
+	var enqueueRequest *enqueueAPIRequest
+	decoder := json.NewDecoder(r.Body)
+	err := decoder.Decode(&enqueueRequest)
+	if err != nil {
+		p.API.LogError("Unable to decode JSON err=" + err.Error())
+		w.WriteHeader(http.StatusBadRequest)
```
levb: This may already be the convention in the TODO plugin, but generally we should return JSON errors.

server/store.go
```diff
@@ -125,8 +157,20 @@ func (p *Plugin) getItemsForUser(userID string) ([]*Item, error) {
 }
 
 func (p *Plugin) addToOrderForUser(userID string, itemID string) error {
```
levb: s/Order/Own/ or whatever the name ends up?

cpoile:
@larkox let me know when you're ready for a review!

levb:
@larkox this will need a PM/UX review - do you mind including screenshots of the new functionality here to make it easier?

server/store.go
```diff
@@ -216,41 +214,41 @@ func (p *Plugin) popFromOrderForUser(userID string) error {
...
 	if appErr != nil {
 		return false, errors.New(appErr.Error())
 	}
 
 	return ok, nil
 }
 
-func (p *Plugin) getOrderForUser(userID string) ([]string, []byte, error) {
-	originalJSONOrder, err := p.API.KVGet(getOrderKey(userID))
+func (p *Plugin) getListForUser(userID string, listID string) ([]*OrderElement, []byte, error) {
```
levb: 1/5 s/getListForUser/getList/

server/store.go
```diff
@@ -216,41 +214,41 @@ func (p *Plugin) popFromOrderForUser(userID string) error {
 	return errors.New("unable to store order")
 }
 
-func (p *Plugin) storeOrder(userID string, order []string, originalJSONOrder []byte) (bool, error) {
+func (p *Plugin) storeList(userID string, listID string, order []*OrderElement, originalJSONOrder []byte) (bool, error) {
```
levb: 1/5 s/storeList/saveList/ - I just learned myself that Get/Sabe is the de-facto convention in the mattermost code base.

server/plugin.go
```diff
@@ -138,6 +156,123 @@ func (p *Plugin) handleList(w http.ResponseWriter, r *http.Request) {
 	w.Write(itemsJSON)
 }
 
+type enqueueAPIRequest struct {
```
levb: 2/5 I _really_ don't like the term `Enqueue` here, can we express it with the existing vocabulary?

server/command.go
```diff
@@ -85,20 +99,64 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+
+	appErr := senderList.add(senderItem.ID, receiverItem.ID, receiver.Id)
+	if appErr != nil {
+		return nil, false, appErr
+	}
+
+	appErr = receiverList.add(receiverItem.ID, senderItem.ID, extra.UserId)
+	if appErr != nil {
+		return nil, false, appErr
+	}
```
cpoile: There's a few things here that we should work on, but it will take a while to explain. Bear with me. Please don't take this as criticism of your ability at all, it's just things we all have to learn -- imagine everything is bracketed with "this is something you learn after working with Go and how we write plugins". :)

1. If you're creating a new item,  the `command.go` doesn't need to know about everything after that. Its job is to call the APIs and get things moving, but it doesn't need to know what each user has in their list, and it shouldn't be inserting todos into lists itself.
1. The List manager should probably be an API that the command calls, and the API will return an error if it failed, or nothing otherwise. something like `listManager.Add(toUserId, fromUserId, message)`.
1. The List manager will do all the "business logic" of taking a laded list (from the store), inserting todos into lists, and then telling the store to save the list. It should call the store. 
1. The store know how to load and save lists, it shouldn't know anything about what goes in the list, or how to manipulate the lists. It should have implement the `ListStore` interface, which is defined in the List Manager's file (because the List Manager is the struct that will use the ListStore to load/save lists).
1. The plugin is in charge of wiring these together. It will call NewListStore (passing in the plugin.API and maybe the plugin.Helpers) to get an instance of the store. It will pass this store into the List Manager, along with anything else the List Manager needs.
1. All of the queueing/dequeueing etc. should happen in the List Manager -- the plugin doesn't need to know about that.

I realize a lot of this design was done before you touched it, but it's good to improve the design as we go. :) 

I'm seeing if I can give you access to one of our repos that gives a good example of this.

server/list.go
```diff
@@ -0,0 +1,78 @@
...
+	InListKey = "_in"
+	// OutListKey is the key used to store the order of sent todos
+	OutListKey = "_out"
+)
+
+func (p *Plugin) getInListForUser(userID string) *list {
+	return &list{
+		listID: InListKey,
+		userID: userID,
+		p:      p,
```
cpoile: In addition to the above comments:

1. Once you refactor, you shouldn't need to put the plugin in the list struct. Generally, think: does the list struct need to know about the plugin? I would say no.
1. The list manager will do all the getting, adding, removing, and have `getItems` and `getItemsOrdered` functions. 

We'll work on more specifics once you've gone through some refactoring. :)

larkox:
I just pushed some commits with changes on the front-end which led to some changes on back-end.

I consider right now the RHS feature complete for this PR. The only missing piece so far is the is the "bump" or "remind" button, but due to the size of the task, maybe it is better to leave it for a later PR. Nevertheless, if you want it on this one, I can squeeze it into ti.

What would be missing for the RHS (apart of code refactoring probably) is the design part (colours, icons, button locations...). This is the current look:

My List:
image

Inbox:
image

Outbox
image

There is one important difference with my original design. The items when they are completed or deleted they are removed from both users. Nevertheless, they will receive a bot message with who deleted it, how, and what was the content of the todo.

Remaining parts on the front-end would be improving the bots messages and add the Send to option on the "Add new todo" modal. Here is how bots messages look right now:
image

server/command.go
larkox: You are completely right. I was struggling to understand how to do the new architecture (no states, but inferred by their list) and adding the lists to simplify a part of the code that I got into really bad code. Now that I have everything more settled in my head, I think I can do better.

I will try my best to do all the refactoring using your comments. Thanks!

server/plugin.go
larkox: Not sure what you meant here. Can you elaborate?

cpoile:
@larkox very nice! I'll review the front end code after we finish up the backend (it will cause some changes to the front end).

server/plugin.go
larkox: What do you think about "Accept"? Since it is used to "accept as my own to do something I have been sent", I think it should work.

larkox:
Here is a first draft on how the bot DM looks:
image

server/item.go
```diff
@@ -12,7 +14,23 @@ type Item struct {
...
 }
 
-func itemsListToString(items []*Item) string {
+// ExtendedItem extends the information on Item to be used on the front-end
+type ExtendedItem struct {
+	Item
+	ForeignUser     string `json:"user"`
+	ForeignList     string `json:"list"`
+	ForeignPosition int    `json:"position"`
+}
```
cpoile: Nice use of embedding -- but consider just putting it all in the Item struct, since I assume we'll be using these Foreign keys all the time, right? If so, they're just a part of item.

server/list.go
```diff
@@ -0,0 +1,239 @@
...
+	OutListKey = "_out"
+)
+
+// ListStore represents the KVStore operations for lists
+type ListStore interface {
+	AddItem(item *Item) error
+	GetItem(itemID string) (*Item, error)
+	RemoveItem(itemID string) error
+
+	GetItemOrder(userID string, itemID string, listID string) (*OrderElement, int, error)
```
cpoile:
```suggestion
	GetItemOrder(userID, itemID, listID string) (*OrderElement, int, error)
```
You can simplify the parameter list like this.

server/list.go
```diff
@@ -0,0 +1,239 @@
...
+	Add(userID string, itemID string, listID string, foreignUserID string, foreignItemID string) error
+	Remove(userID string, itemID string, listID string) error
+	Pop(userID string, listID string) (*OrderElement, error)
+
+	GetList(userID string, listID string) ([]*OrderElement, error)
+}
+
+type listManager struct {
+	store       ListStore
+	getUserName func(string) string
```
cpoile: I assume you need the `plugin.API`, that's why you're passing in the function, right? You can just store a reference to the `plugin.API` here instead, and then ListManager can either have it's own call to get the userName, or share a utility function. Usually in Go we don't store functions for that reason. (We would store them for other reasons, like a swappable algorithm, aka, the "Strategy" pattern.)

server/list.go
```diff
@@ -0,0 +1,239 @@
...
+	return &listManager{
+		store:       store,
+		getUserName: getUserName,
+	}
+}
+
+func (l *listManager) Add(userID string, message string) error {
+	item := newItem(message)
+
+	l.store.AddItem(item)
```
cpoile: unhandled error (and for all the other uses of `AddItem` below)

server/list.go
```diff
@@ -0,0 +1,239 @@
...
+
+	if err := l.store.Add(userID, item.ID, MyListKey, "", ""); err != nil {
```
cpoile: Why are there two functions on the store for `Add`ing?
Another idea: maybe add comments to the functions -- that's where you can explain the differences (and the meaning of the parameters).

server/list.go
```diff
@@ -0,0 +1,239 @@
...
+		return "", "", false, err
+	}
+
+	l.store.RemoveItem(itemID)
+
+	if oe.ForeignUserID == "" {
+		return "", "", false, nil
+	}
+
+	list, _, _ := l.store.GetItemListAndOrder(oe.ForeignUserID, oe.ForeignItemID)
```
cpoile: I wonder -- can you think of a way to change how the items, list and order are stored? There is probably a simpler way... What do you think? (It might be too much work though, so feel free to say its out of scope.)

server/plugin.go
cpoile: Take a look at how workflow or jira returns errors -- as json objects, so that the front end can decode them and display them to the user in the UI.

server/plugin.go
```diff
@@ -173,3 +282,11 @@ func (p *Plugin) sendRefreshEvent(userID string) {
 		&model.WebsocketBroadcast{UserId: userID},
 	)
 }
+
+func (p *Plugin) getUserName(userID string) string {
+	userName := "Someone"
+	if user, err := p.API.GetUser(userID); err == nil {
```
cpoile: The convention in Go is to test for `err != nil` -- if you don't, some people will misread this and be confused.

server/item.go
larkox: I am slightly against this because ForeignList and ForeignPosition may change during the time, and it is the "foreigner" the one changing those values. In order to avoid problems of conflicts in concurrency (two people modifying the same item at the same time) the items can only be added or removed. They never change during their lifetime.
I can add these fields to simplify the number of types in the code, but we would have to keep in mind that what is stored in the KV store is just a part, and the rest has to be recalculated.
One of the invariant parts at the moment would be the ForeignUser, but if we want in the future to create a "send" option for existing todos, or something similar, it will also change.

server/list.go
larkox: 0/5 I like it better when every argument has its type near. Nevertheless, I can change it.

server/list.go
larkox: I thought passing the API here was a bit of overkill, because we just need a function. Also, I did not want to repeat code redefining "getUserName". Nevertheless, you are right about sharing the utility function. I will do that.

server/list.go
larkox: This is something that is also buggering me. There are two different entities in place here.
First one would be the items. Items are invariable during its lifetime, and just record basic information about the todo: ID, Create time, and Message.
On the other hand, there are the "lists". Lists are stored as a list of order elements. On the previous implementation, this order elements were just the IDs of the items. Now the order elements have more information.
The AddItem is used to store items.
The Add is used to add a new Order Element to a list.
I will add comments to make it more understandable, but maybe there should be a bigger change here. If you have any idea, please let me know.

cpoile: I suppose the best we can do for now is hide this complexity in the manager, and not let it leak into the command runner or the store. This is the "business-logic". 

We should leave it for another PR to refactor how lists and items are stored to simplify this (if that's possible).

server/list.go
cpoile: It's totally fine for the ListManager to hold a ref to the API -- we tend to pass that around whenever we need it.

server/list.go
cpoile: It's more idiomatic to try to reduce the amount of text to read. You get used to it. :)

server/item.go
cpoile: Ok, sounds like a good reason to me. Let's keep it, so that you don't have to keep anything in mind. ("Keeping things in mind" is usually a bad thing.)

server/list.go
```diff
@@ -15,65 +17,89 @@ const (
...
+	}
+
 	receiverItem := newItem(message)
-	l.store.AddItem(receiverItem)
+	if err := l.store.AddItem(receiverItem); err != nil {
+		l.store.RemoveItem(senderItem.ID)
+		return "", err
+	}
 
 	appErr := l.store.Add(senderID, senderItem.ID, OutListKey, receiverID, receiverItem.ID)
```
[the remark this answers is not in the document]
larkox: I know I used to think I had a good reason to use appErr, but now I do not know. Will change it right away.

server/list.go
```diff
@@ -15,65 +17,89 @@ const (
...
 	if appErr != nil {
+		l.store.RemoveItem(senderItem.ID)
```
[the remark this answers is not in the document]
larkox: Here I am not dealing with the errors on purpose. There is nothing to handle if something fails, since this is some kind of "rollback" of the changes. The only think I can do is keep trying to remove the item until it gets erased, but since it may fail for other things, it may not be the best approach.

Another possibility would be to include these IDs to a "Garbage collector" to retry the remove later, but that feels a bit of a overkill.

larkox:
I just added the user selector to the modal as it is right now. Here is a screenshot of how it looks at the moment:
image

larkox:
I just added the bump functionality. It gets a message that is still in the inbox of another user, and bump it to the top of the inbox, and send a message to that person. This is how the out list looks now:
image

server/list.go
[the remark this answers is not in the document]
larkox: I just made some changes regarding this, but I am not sure how good they are. The thing is that many errors are dealt with "silently", as I explained in my previous comment. But it might be interesting to have them logged to see if anything is going wrong. I struggled whether to put them as Error or as Debug, but I ended up setting them as errors.

Also, I do not feel specially comfortable with how the code looks now after the changes regarding this, but right now I cannot think of a better way to log these errors. What do you think?

webapp/src/actions.js
```diff
@@ -139,3 +143,24 @@ export const enqueue = (id) => async (dispatch, getState) => {
...
 };
+
+export const bump = (id) => async (dispatch, getState) => {
+    await fetch(getPluginServerRoute(getState()) + '/bump', {
+        method: 'POST',
+        credentials: 'same-origin',
+        headers: {
+            'Content-Type': 'application/json',
+            'X-Requested-With': 'XMLHttpRequest',
+        },
```
cpoile: I think you get the same-origin and headers for free if import client4 from mattermost-redux (see then client.js file from workflow).

server/list.go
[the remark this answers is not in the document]
cpoile: I think this is fine -- it also helps with documenting what all the calls are supposed to be doing. I found it helpful for that reason alone. :) 

Also, the comments you added to the functions are very helpful as well. Thanks!

webapp/src/actions.js
larkox: Regarding Typescript, I think it would be a bit out of scope, but I can try. Nevertheless, I feel makes more sense a new PR to migrate the whole plugin. Also, the code is mostly from webapp, but I removed some things that felt unneeded for this implementation, or that included too many more components from webapp.

Regarding the headers, solved in the new commit.

server/store.go
```diff
@@ -20,126 +21,146 @@ const (
...
 )
 
-func (p *Plugin) storeLastReminderTimeForUser(userID string) error {
-	strTime := strconv.FormatInt(model.GetMillis(), 10)
-	appErr := p.API.KVSet(getReminderKey(userID), []byte(strTime))
-	if appErr != nil {
-		return errors.New(appErr.Error())
-	}
-	return nil
+// OrderElement denotes every element in any of the lists. Contains the item that refers to,
```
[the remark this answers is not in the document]
levb: After reviewing the code some more, I propose 3 concepts/names:
- A `Task` (aka item)
- A `TaskRef` (OrderElement)
- A `List` (aka Order, `[]*TaskRef`)

It may also be worth considering using the term `Issue` to avoid reinventing the wheel and colliding with "agile" naming convention later (Task vs Story???). I still think that `Item` is a bit too abstract.

server/list.go
```diff
@@ -0,0 +1,342 @@
...
+	// MyListKey is the key used to store the order of the owned todos
+	MyListKey = ""
+	// InListKey is the key used to store the order of received todos
+	InListKey = "_in"
+	// OutListKey is the key used to store the order of sent todos
+	OutListKey = "_out"
+)
+
+// ListStore represents the KVStore operations for lists
+type ListStore interface {
```
levb: 1/5 I am confused about this interface.

It'd appear to me that it should provide basic operations on `List` and `Task` keys, and leave the rest to the higher level API implementation.

server/store.go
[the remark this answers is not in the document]
larkox: Maybe it is a bit redundant, but if we end up calling the plugin "Reminder" wouldn't it be nice to call the items reminders? May collide with the current remind functionality, but we can work around that.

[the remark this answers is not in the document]
larkox: I just realised that if we expect to keep backwards compatibility, using the word reminder will be annoying, since the Reminder key for KV store is "reminder". I will go with Issue then.

server/plugin.go
```diff
@@ -17,6 +18,28 @@ const (
 	WSEventRefresh = "refresh"
 )
 
+// ListManager representes the logic on the lists
+type ListManager interface {
+	// Add adds a todo to userID's myList with the message
+	Add(userID, message string) error
```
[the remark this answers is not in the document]
larkox: Pop performs the action in a more atomic way. If we were to do this, we might try to remove an item that might be removed between the get and the complete.
Also, I have my concerns on whether popping is Complete or Remove.

server/list.go
larkox: From my point of view, the only high level functions are:
- Pop and bump, which are here for atomicity
- GetIssueReference and GetIssueListAndReference which I feel they make things clearer if they are at this level, since they are not so much related with the business logic and more with the data shape.

server/plugin.go
```diff
@@ -59,36 +86,59 @@ func (p *Plugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r *http.Req
...
+		p.handleEnqueue(w, r)
+	case "/bump":
+		p.handleBump(w, r)
 	default:
 		http.NotFound(w, r)
 	}
 }
 
+type addAPIRequest struct {
+	Message string
```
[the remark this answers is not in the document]
larkox: In order to be more informative in the code, we may add json tags, but they are not needed. From [json Go documentation](https://golang.org/pkg/encoding/json/#Unmarshal):

> To unmarshal JSON into a struct, Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match.

Therefore, unless you want to make it more verbose, it should not be needed.

server/list.go
levb: Ok, time will tell - we have the pattern of "index keys" in other plugins, hopefully a best practice of how to package it will materialize, we are a bit all over the place now.

server/plugin.go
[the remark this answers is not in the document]
levb: Probably marginally faster if we use the tags and avoid case-insensitive comparison. I also personally like the tags highlight the JavaScript/JSON nature of the go type, typographically make it stand out in your face. But I'm 0/5 on this. Tags are visual clutter, too.

server/store.go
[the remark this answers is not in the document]
levb: Personally, `Issue` makes me cringe, but I think it might be the most intention-revealing, generally accepted name for the entity.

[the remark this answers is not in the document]
larkox: I have also some problems with issue. Maybe it is because English is not my mother tongue, but whenever I hear Issue I hear Problem, and I do not like transmitting with the todos the feeling of "you have to deal with a problem". Also, I wonder how much it will collide in the mind with other uses of Issue, like GitHub Issues.

Nevertheless, I have no better alternative apart of maybe "Todo" itself, but if we get a new name to the plugin, we may find with it the new name for the element.

levb:
@larkox remove the WIP status? ready for review?

aaronrothschild:
This looks awesome @larkox a few comments:
1) The naming of the tabs in the RHS seem a bit off.  Perhaps "My Todos", "Inbox","Sent".  The use of synonyms that are associated with email organization gives the impression that other users are involved and clarifies in the user's mind what the tabs are all about.  The first "My Todos" shows that this is a list of objects called "Todos" that I'm dealing with.   Then, with that context in mind - I can see there is an "Inbox" which likely involves other users, great.  Oh, cool I can send Todos as well.  3/5 Open to other ideas that help clarify the mental model for the user.  @asaadmahmood 
2) When the tab is open to "Out" - can we change the text "Add New Item" to say "Request a To-do from someone" 
3) Can you send a screenshot of how the "user selector" works when creating a new task for someone?  I'm wondering if you can type to quickly get to a specific user.  I think we may have a pattern for user selection in another plugin project (perhaps a separate PR).  My main concern is in installations with hundreds/thousands of users and making it easy to select someone when there may be multiple "John Smith"'s.

webapp/src/actions.js
cpoile: Yep, don't worry about Typescript for now. It's for another PR.

server/plugin.go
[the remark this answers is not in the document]
cpoile: Personally, I prefer to explicitly have the JSON tags because it reminds you that this is serialized. And as soon as you add a camel-case field you'll need to write the tag to turn it into json-style snake-case, so it reminds you that you need to do that.

larkox:
Thanks for the comments, @aaronrothschild 
1. I actually had them like that on my first version, but I changed them because @levb comments. Maybe he has a stronger opinion against "Inbox" or "Sent", but I am not sure.
2. Would be no problem.
3. It works almost exactly as in interactive dialogs (I copy pasted the code from webapp) which should behave quite similar to user autocomplete on search box or message box. You can write the user name and will get you the relevant users.

aaronrothschild:
> Thanks for the comments, @aaronrothschild
> 
> 1. I actually had them like that on my first version, but I changed them because @levb comments. Maybe he has a stronger opinion against "Inbox" or "Sent", but I am not sure.

I feel relatively strongly about this - I'll wait for @asaadmahmood to comment with his thoughts.  These are easy to change later, but I think these are clearer to the user.  cc @levb 

> 2. Would be no problem.
> 3. It works almost exactly as in interactive dialogs (I copy pasted the code from webapp) which should behave quite similar to user autocomplete on search box or message box. You can write the user name and will get you the relevant users.

Awesome, this is a really great addition to the task app!

levb:
> 1. ... "My Todos", "Inbox","Sent".

WFM, with caveats. 1) can we be short? Isn't "My" enough? 2) The next feature I'd love to file for consideration is sending pings to a channel; the channel then would have a visual indication that there are actionable items (and a poppable RHS) and anyone in the channel can act on them. Would these still be "Sent" from my perspective? I still like My/In/Out as the shortest descriptive, but really 0/5.

> 2. When the tab is open to "Out" - can we change the text "Add New Item" to say "Request a To-do from someone"

0/5 Can it be just "New" and "Request" (pin-n-ng)?

But really, we should do as @aaronrothschild and @asaadmahmood say.

asaadmahmood:
I agree with @aaronrothschild, I didn't really understand at first what in and out were. Were "in" the ones that I have done, or accepted, and "out" the ones that were "outed"?
If we want that in/out terminology, we can use incoming and outgoing, but I feel like that is too complex as well.

Personally, I would like to simplify this even further.
We can just have **Inbox** and **Sent**.

**Inbox** would include my todos, whether they were creating by me, or assigned to me by someone else, or My Todos that are shared with me and another user etc.
**Sent** would include todos that I have sent to others.

@aaronrothschild your comment:
> The first "My Todos" shows that this is a list of objects called "Todos" that I'm dealing with. Then, with that context in mind - I can see there is an "Inbox" which likely involves other users, great.

I don't like the separation of My Todos and Inbox, which are both "My Todos" right. If Inbox includes other users apart from me, then I don't see any reason to separate it.
That would be a UX defect of this model, I would have to switch back and forth on a Todo that concerns me.
If we want, we can make it apparent in the **Inbox** tab which were are shared todos, and which one are only mine, but separating them out into tabs adds additional clicks.

larkox:
@asaadmahmood in Inbox there are only "new todos", in the sense of incoming todos that I have not yet processed. By processing I mean: dismissing it (won't do), confirm that is already done, or accept. When you accept a todo, it moves to your list, and you will see it on "my todos" list. Also, it is no longer removable by the sender.

asaadmahmood:
@larkox Got it, but I'm wondering, why should a user switch contexts to accept/reject new todos, we can just display incoming todos right then and there and a user has the ability to Accept/Reject them. If they reject it, they removed from the Inbox screen, if they accept it, they are added.

Instead of Inbox, we can just have "Todos", and "Sent" then.

That way, a user would not have to accept a Todo in inbox, then jump back again, and then if he sees a new Todo in inbox, jump back again, accept it, and then back to the First tab.

I don't want this back and forth.

aaronrothschild:
> @larkox Got it, but I'm wondering, why should a user switch contexts to accept/reject new todos, we can just display incoming todos right then and there and a user has the ability to Accept/Reject them. If they reject it, they removed from the Inbox screen, if they accept it, they are added.
> 
> Instead of Inbox, we can just have "Todos", and "Sent" then.
> 
> That way, a user would not have to accept a Todo in inbox, then jump back again, and then if he sees a new Todo in inbox, jump back again, accept it, and then back to the First tab.
> 
> I don't want this back and forth.

I think this is a good approach, I like the reduced switching and the reduced number of tabs too.

larkox:
Right now, the new todos are always at the bottom of the list. If the user has a long list of todos, may not see the new ones, and therefore, never informing the sender that I have accepted that todo. This could be solved with the bot message that is already in place, which also has interactive buttons, but I still feel that a constant reminder of "you have not answered to this" might be necessary.

I could propose to keep unanswered messages on top, but I am not sure if that is the best approach.

Nevertheless, take in mind that all this comments are with a 0/5. I am probably polluted by the current architecture.

asaadmahmood:
Incoming todos should always show on the top, however, as they can actually be treated like mentions.
However, we can have a divider like, that allows users to collapse it.
This way, we can have both sort of functionalities in the same place.
If users are already caught up, they can keep the Incoming todo list expanded.
If they have tons of Todos in the incoming todos list, and want to deal with it later, they can just collapse it and work on their personal ones.

Incoming Todos (3) --------------- (collapse)
Todo 1                                       (Accept - Reject)
Todo 2                                       (Accept - Reject)
Todo 3                                       (Accept - Reject)

My Todos (2) ---------------------- (collapse)
Todo 1                                       (Done - Remove)
Todo 2                                       (Done - Remove)

levb:
@asaadmahmood makes perfect sense, I 100% defer

larkox:
Here is the new behaviour:
image

As you can see, there is the inbox list (Incoming Todos) and the my list (My Todos). On top, I was not sure how to deal with the numbering. Right now is "Todos" + number of my todos + number of inbox todos. Nevertheless, does not feel clear. I also thought the possibility of only notifying the inbox todos, but not sure if that is enough information. Another last idea was to notify the sum of inbox and my. Which one do you prefer? Do you have better ideas?

Also, both list can be collapsed:
image

image

image

Finally, as you can see, the style is not the prettiest. Any suggestion on style would be much appreciated.

larkox:
Here also some edge cases.

No my elements:
image

No inbox elements:
image

levb:
@larkox can you please resolve the conflicts?

aaronrothschild:
@asaadmahmood Could you give us a hand with the UX/styling of this?

levb:
> @asaadmahmood Could you give us a hand with the UX/styling of this?

@aaronrothschild @larkox I'm motivated to get this in and wrap up the work on the ticket. Any chance we should consider the UX/styling as a separate ticket and get this out first, or is it not yet acceptable for community.mattermost.com?

Also, @cpoile - are there any remaining items for this PR or can you approve from your end?

larkox:
I can try to make it less horrible to the sight, but with proper guidance it will be faster and with less repetition. If getting that guidance will take too long and you want it out, I have no problem on investing some mana into to give it a "not so bad" appearance.

cpoile:
Sorry, thought I had approved already. :)

larkox:
I slightly improved the style of this PR so we are able to merge the current version.
Here a few screenshots:
Todo message from bot:
image
My Todos list expanded:
image
My Todos list collapsed:
image
Sent todos:
image

aaronrothschild:
Looks better than before!  ;) @larkox Looking forward to polishing this off later, but will be good to get some feedback on the new functionality.

levb:
@chrisp already approved and it's been all stylistic after; @DHaussermann it's ready for QA.

DHaussermann:
There some issues with this. Maybe they can't all be fixed on this PR? Let me know how is best to organize the work. 

1. I can select myself as the user I want to send a task to. It then shows in my sent list (as well as my own list) and I can bump myself about it or cancel
2. When you send a task to a user and they set it to done (either without adding to their list or after) - The DM received from the bot by the sender incorrectly states that the sender completed the task and not the user who actually set the task as done.  
WrongUsercompletedTask
3. Lack of profile images in user auto-complete is off putting as everywhere else in the UI shows the profile images so, this looks broken. 
4. When a task is accepted by the recipient. It shows `Accepted. On position 0` I'm guessing this should be the date the task was accepted on?
NoAcceptedDateStamp
5. Horizontal text alignment issue for incoming To Dos
TextAlignment
6. Buttons need UX styling when LHS is expanded. They are awkwardly wide. I can't think of an existing model for what button should look like when LHS is expanded.
ButtonsWideLHS
7. Button color is white on mouse over with white text - MacOs only 
WhiteOnWhite

DHaussermann:
@larkox Sorry, I forgot to mention you in the comment above. Can you please let me know which of the issues above can be addressed in the scope of this PR?

larkox:
> I can select myself as the user I want to send a task to. It then shows in my sent list (as well as my own list) and I can bump myself about it or cancel

Selecting yourself as a user is allowed. Having it on your sent list is a problem. I will look into it.

> When you send a task to a user and they set it to done (either without adding to their list or after) - The DM received from the bot by the sender incorrectly states that the sender completed the task and not the user who actually set the task as done.

Will fix.

> Lack of profile images in user auto-complete is off putting as everywhere else in the UI shows the profile images so, this looks broken.

This is hard to fix. The selector is not exactly the same as in the webapp, and bringing all the components and logic to show these will be quite troublesome. @aaronrothschild what do you think about leaving it like this for the time being?

> When a task is accepted by the recipient. It shows Accepted. On position 0 I'm guessing this should be the date the task was accepted on?

It is the position on their Todo list. Probably I should change that to be 1 instead of 0 (programmer things...). What do you think? Or would you prefer a different wording?

> Horizontal text alignment issue for incoming To Dos

I may be blind, but I do not see where the problem is.

> Buttons need UX styling when LHS is expanded. They are awkwardly wide. I can't think of an existing model for what button should look like when LHS is expanded.

Not sure how much will it take at the moment, and may change once we have a better understanding on how we want this to look. @aaronrothschild are you OK if we leave it as it is for now?

> Button color is white on mouse over with white text - MacOs only. (Chrome and desktop)

This is weird, and should be a problem with CSS. @aaronrothschild should I investigate further, or shall we solve it later?

larkox:
@DHaussermann I also forgot to mention you...

asaadmahmood:
@aaronrothschild Yup, I think we can take care of the UI improvements later, I won't be able to help much right now, as I'm busy on a few other things on my plate.

DHaussermann:
5. Is very minor but there is a bit of white space missing. The text begins very close to the edge. I updated the screenshot. Definitely not a high priority issue.

larkox:
@DHaussermann Fixed 1, 2, 5 and improved slightly 4.

aaronrothschild:
@larkox What is the improvement in #4 ?  We can definitely ship on this and improve on the other items afterwards.  #iteration

larkox:
@aaronrothschild Now I start on 1, instead of on 0.

DHaussermann:
Thanks @larkox for the fixes!

1. --> Confirmed fixed 
2. --> Confirmed fixed
5. --> Confirmed fixed 

4. For the position text, the counting does work now. It starts at 1. 
My issue with it was not about how developers start counting at 0 instead of 1. I knew that should be easy to fix :)
I mean more broadly - I think it looks odd and it's a bit confusing. It needs a bit of UX polish but also, until position can be modified by the user, (prioritization of tasks - Issue here https://github.com/mattermost/mattermost-plugin-todo/issues/33 ) knowing the position is not meaningful since it's just the order the task were accepted in and not the priority. But, I appreciate that while it confusing for now, it will make more sense when tasks can be prioritized by the user.   
cc @aaronrothschild 

I also found a  issue that was not reproducing yesterday.

8. _Add to To Do_ button text is no longer readable. The default state is white text on an light gray button. 
AddButton
@larkox can you please see if you can repro this ☝️ 

One other question - Why is there no list item for this in the System Console on the left hand side of the admin menu - under the plugin section. Is this typical when there is nothing to configure?

larkox:
@DHaussermann The button is like that as long as there is no message on the textbox (as expected). If you are seeing this behaviour in any other case, I am not sure what could be the root.

DHaussermann:
@larkox I understand that the button is in a disabled state when the box above has no text. My issue is that even in the disabled state, the text should be readable. Can we darken the color to make it more readable or does that require CSS changes that will impact other UI elements?

larkox:
@DHaussermann I put the text black. This is also something I did with other buttons, but this will ignore the theme.

larkox:
@DHaussermann Solved!

Issue #25, opened 2020-01-15
"[x] add to thread" mode


levb:
1) When creating a new TODO from a post (in the modal) add a checkbox [x] add this item to the conversation (thread?).
2. When the item is created, post a message to the thread - "Marked as Todo for @user"
3. When the item is popped, post a message to the thread "@user resolved ..."

cc @jwilander @aaronrothschild for further suggestions/edits.

larkox:
May I take this one?

jwilander:
Go for it @larkox !

larkox:
I have a good idea on how to do this, but the changes on the architecture with [this PR](https://github.com/mattermost/mattermost-plugin-todo/pull/27) will prove quite useful for dealing with this beautifully. Therefore, I would prefer to wait for that PR to be merged.

Regarding the solution, the idea would be to add the PostID to the Issue object. and whenever the issue is removed, if there is a postID, reply to the post.

jwilander:
That sounds reasonable to me @larkox 👍

PR #40, opened 2020-03-20
Add todo to a thread

larkox:
#### Summary
Now you can add a todo to a Thread. When you open the Add to do from the dot menu of a post, you will see now an option to add the todo to the thread.

Screenshot from 2020-03-20 18-55-22

When you create the todo, a message will be sent to the thread, and another message when the todo is deleted (by completing, cancelling, or "won't do"-ing it).

Screenshot from 2020-03-20 18-57-17

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-todo/issues/25

### DIFF
--- a/server/bot.go
+++ b/server/bot.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/pkg/errors"
 )
 
 // PostBotDM posts a DM as the cloud bot user.
@@ -49,4 +51,34 @@ func (p *Plugin) PostBotCustomDM(userID string, message string, todo string, iss
 	})
 
 	return appError
+}
+
+// ReplyPostBot post a message and a todo in the same thread as the post postID
+func (p *Plugin) ReplyPostBot(postID, message, todo string) error {
+	if postID == "" {
+		return errors.New("Post ID not defined")
+	}
+
+	post, appErr := p.API.GetPost(postID)
+	if appErr != nil {
+		return appErr
+	}
+	rootID := post.Id
+	if post.RootId != "" {
+		rootID = post.RootId
+	}
+
+	quotedTodo := "\n> " + strings.Join(strings.Split(todo, "\n"), "\n> ")
+	_, appErr = p.API.CreatePost(&model.Post{
+		UserId:    p.BotUserID,
+		ChannelId: post.ChannelId,
+		Message:   message + quotedTodo,
+		RootId:    rootID,
+	})
+
+	if appErr != nil {
+		return appErr
+	}
+
+	return nil
 }
--- a/server/command.go
+++ b/server/command.go
@@ -118,7 +118,7 @@ func (p *Plugin) runSendCommand(args []string, extra *model.CommandArgs) (*model
 
 	message := strings.Join(args[1:], " ")
 
-	receiverIssueID, err := p.listManager.SendIssue(extra.UserId, receiver.Id, message)
+	receiverIssueID, err := p.listManager.SendIssue(extra.UserId, receiver.Id, message, "")
 	if err != nil {
 		return nil, false, err
 	}
@@ -143,7 +143,7 @@ func (p *Plugin) runAddCommand(args []string, extra *model.CommandArgs) (*model.
 		return getCommandResponse(model.COMMAND_RESPONSE_TYPE_EPHEMERAL, "Please add a task."), false, nil
 	}
 
-	if err := p.listManager.AddIssue(extra.UserId, message); err != nil {
+	if err := p.listManager.AddIssue(extra.UserId, message, ""); err != nil {
 		return nil, false, err
 	}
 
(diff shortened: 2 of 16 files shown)


server/plugin.go
```diff
@@ -21,19 +21,19 @@ const (
 // ListManager representes the logic on the lists
 type ListManager interface {
 	// AddIssue adds a todo to userID's myList with the message
-	AddIssue(userID, message string) error
+	AddIssue(userID, message, postID string) error
```
jfrerich: When postID added to the issue struct, let's pass and consume the `issue` object in methods.

jfrerich:
@larkox, that last commit looks nice.  Thanks for the edits.

I have one more item

image

Regarding the UI/UX for the new option, I feel a bit more description in the modal is necessary.  When the `add to thread` option is selected, the actual translation is "have the todo bot post updates to the thread when the todo issue has changed."

1/5 I propose replacing the current text with 
- [ ] **Add to thread** *Select to have the Todo Bot respond to the thread when the attached todo is added, modified, or completed.*

Thoughts?

server/bot.go
```diff
@@ -50,3 +52,33 @@ func (p *Plugin) PostBotCustomDM(userID string, message string, todo string, iss
...
+	if post.RootId != "" {
+		rootID = post.RootId
+	}
+
+	quotedTodo := "\n> " + strings.Join(strings.Split(todo, "\n"), "\n> ")
+	_, appErr = p.API.CreatePost(&model.Post{
+		UserId:    p.BotUserID,
+		ChannelId: post.ChannelId,
+		Message:   message + quotedTodo,
+		RootId:    rootID,
```
levb: Sorry for my own continued confusion, but should we also be setting the ParentID on this post?

server/store.go
```diff
@@ -86,13 +86,18 @@ func (l *listStore) GetIssue(issueID string) (*Issue, error) {
 	return issue, nil
 }
 
-func (l *listStore) RemoveIssue(issueID string) error {
+func (l *listStore) RemoveIssue(issueID string) (*Issue, error) {
+	issue, err := l.GetIssue(issueID)
```
levb: nit: I question this pattern (a mandatory Get before removing) but only because of the name of this function. Not sure how to name it better though.

levb: Looking through its use, I think you should have 2 functions - one with a get, and one without. Or maybe just `Get` explicitly where needed? Seems like a handful of places.

server/command.go
```diff
@@ -193,23 +193,26 @@ func (p *Plugin) runListCommand(args []string, extra *model.CommandArgs) (*model
...
+		message := fmt.Sprintf("@%s popped a Todo you sent: %s", userName, issue.Message)
+		p.sendRefreshEvent(issue.ForeignUser)
+		p.PostBotDM(issue.ForeignUser, message)
 	}
 
 	p.sendRefreshEvent(extra.UserId)
 
 	responseMessage := "Removed top to do."
 
+	replyMessage := fmt.Sprintf("@%s popped a todo attached to this thread", userName)
```
levb: nit: What's the right format, `todo` or `Todo` (or `TODO`?)

server/list.go
```diff
@@ -56,15 +56,15 @@ func NewListManager(api plugin.API) *listManager {
...
+func (l *listManager) AddIssue(userID, message, postID string) error {
+	issue := newIssue(message, postID)
 
 	if err := l.store.AddIssue(issue); err != nil {
 		return err
 	}
 
 	if err := l.store.AddReference(userID, issue.ID, MyListKey, "", ""); err != nil {
-		if rollbackError := l.store.RemoveIssue(issue.ID); rollbackError != nil {
+		if _, rollbackError := l.store.RemoveIssue(issue.ID); rollbackError != nil {
```
levb: unneeded `Get`...

server/bot.go
larkox: After checking the code and asking in the [Server channel](https://community.mattermost.com/core/pl/uezrbne3utg6bxq9c365wbmndy), RootID and ParentID should have the same ID, and the CreatePost API makes sure that if no ParentID is given, use RootID. If you want I can be verbose here and set parentID to rootID too, but I think it is not needed.

larkox:
@jfrerich Made the change you requested. Here is the look.
Screenshot from 2020-03-24 13-17-03

server/command.go
larkox: The other comments are addressed. I am asking on the todo plugin channel which option looks better, and I will substitute every mention with the final decision.

jfrerich:
@larkox I like it, but will defer to @aaronrothschild for the final decision.

I also think swapping the order makes the UI look nicer, by having the longer texts grouped together.

image

larkox:
@jfrerich I like it better with your change so I made it while we are waiting for Aaron comment.

@levb All your comments should be addressed.

hanzei:
@aaronrothschild Would you be fine if this PR get's merged without your approval as it's blocking the next release?

PR #52, opened 2020-04-20
Use bot account to post slash command responses

attilamolnar:
#### Summary
Use bot account to post slash command responses

#### Ticket Link
#4

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -50,13 +50,13 @@ func getCommand() *model.Command {
 	}
 }
 
-func getCommandResponse(responseType, text string) *model.CommandResponse {
-	return &model.CommandResponse{
-		ResponseType: responseType,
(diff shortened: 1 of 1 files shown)


attilamolnar:
A slightly less intrusive way would be to do [like the gitlab plugin](https://mattermost/mattermost-plugin-gitlab/blob/43a23dc87436d8ecfded81538cffd4ebbb3c39b0/server/command.go#L49-L61), i.e. call `SendEphemeralPost()` from `getCommandResponse()` then return the dummy `CommandResponse`.

Diff for this alternative: https://github.com/mattermost/mattermost-plugin-todo/compare/master...attilamolnar:bot-cmd-response2

hanzei:
Great work @attilamolnar :tada:

Issue #34, opened 2020-03-04
List does not stay current if changes are made while it's open in multiple clients


jfrerich:
### Summary
If the To Do List is expanded in more than one client and edits are made, a user must close and re-open the list to see the current state in other clients.

**Steps**:

* Login to the same server from desktop and a browser (or 2 different browsers)
* Open the To Do plugin in both sessions
* From 1 session add or remove items to your To Do list

**Expected** To Do list is updated in real-time for the other session

### Additional Information

The suggested fix is to should understand/find what websocket notification is missing, and implement it.

### Ticket Link

https://mattermost.atlassian.net/browse/MM-22575

---
If you're interested please comment here and come [join our "Contributors" community channel](https://community.mattermost.com/core/channels/tickets) on our daily build server, where you can discuss questions with community members and the Mattermost core team.

New contributors please see our [Developer's Guide](https://developers.mattermost.com/contribute/getting-started/) and our [Plugins Guide](https://developers.mattermost.com/extend/plugins/).

TheoVitkovskiy:
I will give my best shot on the weekend! (I have a long weekend starting tomorrow)

jfrerich:
Thank you, @TheoVitkovskiy.  Let us know if you need any help!

PR #54, opened 2020-04-21
Add todo button on team sidebar

Ashniu123:
Resolves issue #24

#### Summary
Add 4 buttons on team sidebar as per issue
Buttons are - Your todos, requested todos, assigned todos, refresh

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-todo/issues/24

### DIFF
--- a/webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
+++ b/webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
@@ -30,17 +30,19 @@ export default class SidebarButtons extends React.PureComponent {
         };
     }
 
-    componentDidUpdate() {
-        this.getData();
+    componentDidUpdate(prevProps) {
+        if (prevProps.issues !== this.props.issues || prevProps.inIssues !== this.props.inIssues || prevProps.outIssues !== this.props.outIssues) {
+            this.getData();
+        }
     }
 
     getData = async (e) => {
-        if (this.state.refreshing) {
-            return;
-        }
-
         if (e) {
             e.preventDefault();
+        }
+
+        if (this.state.refreshing) {
+            return;
         }
 
         this.setState({refreshing: true});
@@ -78,41 +80,41 @@ export default class SidebarButtons extends React.PureComponent {
         return (
             <div style={container}>
                 <OverlayTrigger
-                    key='todoLink'
+                    key='myTodosLink'
                     placement={placement}
-                    overlay={<Tooltip id='todoTooltip'>{'Your todos'}</Tooltip>}
+                    overlay={<Tooltip id='myTodosTooltip'>{'Your Todos'}</Tooltip>}
                 >
                     <a
                         style={button}
-                        onClick={() => this.openRHS(RHSStates.ISSUES)}
+                        onClick={() => this.openRHS(RHSStates.InListName)}
                     >
-                        <i className='fa fa-check'/>
+                        <i className='icon icon-check'/>
                         {' ' + issues.length}
                     </a>
                 </OverlayTrigger>
                 <OverlayTrigger
-                    key='todoReviewsLink'
+                    key='incomingTodosLink'
                     placement={placement}
-                    overlay={<Tooltip id='reviewTooltip'>{'Todos received'}</Tooltip>}
+                    overlay={<Tooltip id='incomingTodosTooltip'>{'Incoming Todos'}</Tooltip>}
                 >
                     <a
-                        onClick={() => this.openRHS(RHSStates.IN_ISSUES)}
+                        onClick={() => this.openRHS(RHSStates.InListName)}
                         style={button}
                     >
-                        <i className='fa fa-angle-double-down'/>
+                        <i className='icon icon-arrow-down'/>
                         {' ' + inIssues.length}
                     </a>
                 </OverlayTrigger>
                 <OverlayTrigger
-                    key='todoAssignmentsLink'
+                    key='outgoingTodosLink'
                     placement={placement}
-                    overlay={<Tooltip id='reviewTooltip'>{'Todos sent'}</Tooltip>}
+                    overlay={<Tooltip id='outgoingTodosTooltip'>{'Outgoing Todos'}</Tooltip>}
                 >
                     <a
-                        onClick={() => this.openRHS(RHSStates.OUT_ISSUES)}
+                        onClick={() => this.openRHS(RHSStates.OutListName)}
                         style={button}
                     >
-                        <i className='fa fa-angle-double-up'/>
+                        <i className='icon icon-arrow-up'/>
                         {' ' + outIssues.length}
                     </a>
                 </OverlayTrigger>
@@ -126,7 +128,7 @@ export default class SidebarButtons extends React.PureComponent {
                         style={button}
                         onClick={this.getData}
                     >
-                        <i className={'fa fa-refresh' + refreshClass}/>
+                        <i className={'icon icon-sync' + refreshClass}/>
                     </a>
                 </OverlayTrigger>
             </div>
--- a/webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
+++ b/webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
@@ -30,30 +30,6 @@ export default class SidebarButtons extends React.PureComponent {
         };
     }
 
-    componentDidUpdate(prevProps) {
-        if (prevProps.issues !== this.props.issues || prevProps.inIssues !== this.props.inIssues || prevProps.outIssues !== this.props.outIssues) {
-            this.getData();
-        }
-    }
-
-    getData = async (e) => {
-        if (e) {
-            e.preventDefault();
-        }
-
-        if (this.state.refreshing) {
-            return;
-        }
-
-        this.setState({refreshing: true});
-        await Promise.all([
-            this.props.actions.list(false, 'my'),
-            this.props.actions.list(false, 'in'),
-            this.props.actions.list(false, 'out'),
-        ]);
-        this.setState({refreshing: false});
-    }
-
     openRHS = (rhsState) => {
         this.props.actions.updateRhsState(rhsState);
         this.props.showRHSPlugin();
--- a/webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
+++ b/webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
@@ -1,0 +1,158 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {Tooltip, OverlayTrigger} from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import {makeStyleFromTheme, changeOpacity} from 'mattermost-redux/utils/theme_utils';
+
+import {RHSStates} from '../../constants';
+
+export default class SidebarButtons extends React.PureComponent {
+    static propTypes = {
+        theme: PropTypes.object.isRequired,
+        isTeamSidebar: PropTypes.bool,
+        showRHSPlugin: PropTypes.func.isRequired,
+        actions: PropTypes.shape({
+            list: PropTypes.func.isRequired,
+            updateRhsState: PropTypes.func.isRequired,
+        }).isRequired,
+    };
+
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            refreshing: false,
+        };
+    }
+
+    componentDidUpdate() {
+        this.getData();
+    }
+
+    getData = async (e) => {
+        if (this.state.refreshing) {
+            return;
+        }
+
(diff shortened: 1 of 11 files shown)


Ashniu123:
image
The sidebar looks like this, tell me if any changes are required

larkox:
@asaadmahmood I asked for your review so you can comment on the selection of icons for this ticket. Thanks!

webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
```diff
@@ -0,0 +1,161 @@
...
+        super(props);
+
+        this.state = {
+            refreshing: false,
+        };
+    }
+
+    componentDidUpdate() {
+        this.getData();
+    }
```
larkox, line 35: Why this? My first thought regarding this is that when we get the data, as we finish, we change the refreshing back to false. This will create a new render, which will call this componentDidUpdate, which will call get data, which will change back the state to true, making a new rendering.
Maybe I am missing something, but does look like a dangerous loop.

webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
```diff
@@ -0,0 +1,161 @@
...
+
+    getData = async (e) => {
+        if (this.state.refreshing) {
+            return;
+        }
+
+        if (e) {
+            e.preventDefault();
+        }
```
larkox, line 44: Shouldn't the prevent default be before the early return?

webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
```diff
@@ -0,0 +1,161 @@
...
+        const inIssues = this.props.inIssues || [];
+        const outIssues = this.props.outIssues || [];
+        const refreshClass = this.state.refreshing ? ' fa-spin' : '';
+
+        return (
+            <div style={container}>
+                <OverlayTrigger
+                    key='todoLink'
+                    placement={placement}
+                    overlay={<Tooltip id='todoTooltip'>{'Your todos'}</Tooltip>}
```
larkox, line 83: We try to use on all the user facing strings the term "Todo" (or "Todos" in this case), that is, with the first letter uppercase. Could you correct it here?

webapp/src/reducer.js
```diff
@@ -60,11 +60,21 @@ function rhsPluginAction(state = null, action) {
...
+    }
+}
+
 export default combineReducers({
     rootModalVisible,
     postID,
     issues,
     inIssues,
     outIssues,
+    rhsState,
```
larkox, line 78: I still do not see how this state changes the behaviour on the RHS. Am I missing anything?

Also, take in mind that the RHS only have two different tabs (my Todos and sent Todos). Incoming Todos are in the "my Todos" tab. A nice solution could be to use this state to collapse the incoming Todos or the rest of Todos depending on it.

asaadmahmood:
@Ashniu123 Can you try using mattermost icons in order (we are trying to stray away from font awesome now).
Your Todos: ` icon-check`
Requested/Outgoing: `icon-arrow-up`
Assigned/Incoming: `icon-arrow-down`
Refresh: ` icon-sync`.

You can use these icons like:
`<i className='icon icon-sync'/>`

webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
Ashniu123: You're right! I will add check with `prevProps` and `this.props` to resolve this.

webapp/src/reducer.js
Ashniu123: I intended to open the `Sent` part of RHS when I click on `Todos Outgoing` sidebar button. Do you require it or shall I remove it. It will require some more changes to state and will modify the RHS component to get data from redux.

Ashniu123: Nevermind, I have done it.
On clicking `Your Todos` or `Incoming Todos`, you see the `Todos` tab
On clicking `Outgoing Todos`, you see the `Sent` tab

Ashniu123: @larkox can you confirm if the changes I made are acceptable?

webapp/src/components/sidebar_buttons/sidebar_buttons.jsx
```diff
@@ -0,0 +1,163 @@
...
+        this.state = {
+            refreshing: false,
+        };
+    }
+
+    componentDidUpdate(prevProps) {
+        if (prevProps.issues !== this.props.issues || prevProps.inIssues !== this.props.inIssues || prevProps.outIssues !== this.props.outIssues) {
+            this.getData();
+        }
+    }
```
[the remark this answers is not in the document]
Ashniu123: Yeah. You are right. I saw this thing in the "reference" part of the issue pointing to the github plugin. I did have the same doubt but I thought it was some pattern you guys followed. I have removed it now. Thanks.

Ashniu123:
@larkox can you confirm the changes now? I believe the PR is ready to be merged.

asaadmahmood:
@Ashniu123 Can I get an updated screenshot of how this looks?

Ashniu123:
Yeah Sure. Here it is...

image

aaronrothschild:
> Looks good, but before approving @aaronrothschild, are we sure we're not going to get pushback if we introduce more icons to the LHS?

Agreed.  This is on community and there is more sensitivity to this type of real-estate.  Particularly when the plugin is applied to ALL users of the system, not just users who have "turned on" this plugin (like in Jira, the icons are not visible, until you've linked your account).  

This needs to be an option set by the user for the todo plugin.  The default should be OFF.  set by something like `/todo settings icon-display ON`

aaronrothschild:
@michaelgamble Any thoughts on this or comments re: icons above?  This will eventually end up on community (gated by a setting).

Ashniu123:
I have another concern actually, it is regarding the UI - the "refresh" button.

Other plugins use a "refresh" button to update the data too right? If I had 2 plugins which show a refresh button on LHS - say github and todo. Isn't that a bad UX?

aaronrothschild:
> I have another concern actually, it is regarding the UI - the "refresh" button.
> 
> Other plugins use a "refresh" button to update the data too right? If I had 2 plugins which show a refresh button on LHS - say github and todo. Isn't that a bad UX?

Oh, that's a really interesting point.  I think each plugin implements it's own "refresh" mechanism, which then refreshes the respective icons.  But I see what you're saying - which is that it would be better if there was a single "refresh" button for a user, which would then update all the various plugins... thoughts @levb @asaadmahmood ?

asaadmahmood:
@aaronrothschild I'm like, do we even need a manual refresh button, in most cases I'm assuming it would fetch the data, while in other cases, we can just have a trigger that if a person hovers over the plugin area, we fetch the changes? (This may look weird if the changes only get fetched when the user hovers the plugin area, and is not already fetching for changes intermittently).

levb:
@mickmister @larkox do we have a hook in the client we could use to replace the refresh button with transparent refreshing? Also, why not use websocket push notifications?

mickmister:
@levb The GitHub plugin does a number of things to keep data up-to-date, in addition to having the refresh button.

It makes use of websocket events, browser inactivity, and browser reconnection after some time of disconnect. At the very least, we can implement the websocket strategy, as all transactions regarding the plugin are in-house. I say we make a HW ticket to mirror what GitHub is doing.

https://github.com/mattermost/mattermost-plugin-github/blob/119ff1807a40eca316e5c1644589f056baa71936/webapp/src/index.js#L34-L47

levb:
> websocket strategy, as all transactions regarding the plugin are in-house

@mickmister precisely. should be good enough.

So we have 2 options IMO:
(a) implement websocket updates now, with this PR. @Ashniu123 would you be up for it?
(b) merge this with a refresh button, and log a separate issue to replace it with websockets.

levb:
BTW, this is SO GREAT. I can't wait, button or not! @Ashniu123 thank you for the contribution!

Ashniu123:
The websocket strategy sounds good. Will update this PR soon.

mickmister:
@Ashniu123

> There is a refresh mechanism in place already.

That's great news!

I have one request to include a "reconnect handler", which will run the handler when a browser reconnects its websockets after being disconnected somehow. Here's an example in the GitHub plugin:

https://github.com/mattermost/mattermost-plugin-github/blob/119ff1807a40eca316e5c1644589f056baa71936/webapp/src/index.js#L37

@Ashniu123 What do you think about this?

Ashniu123:
@mickmister - I believe that there won't be a need for "reconnect handler" as such because the "Todo plugin" is internal. That is, it doesn't connect to an external API like Github's which required a "connected" action.

[Github plugin's "connected" action](https://github.com/mattermost/mattermost-plugin-github/blob/119ff1807a40eca316e5c1644589f056baa71936/webapp/src/actions/index.js#L13).

Do correct me if I am wrong.

I shall just remove the refresh button for now.

Ashniu123:
image

The sidebar looks like this now without the refresh button.

mickmister:
@Ashniu123 The purpose of [`registerReconnectHandler`](https://developers.mattermost.com/extend/plugins/webapp/reference/#registerReconnectHandler) actually has nothing to do with the GitHub plugin itself. The function named `getConnected` is confusing in this case, as its name is actually unrelated to the  `registerReconnectHandler` function. `registerReconnectHandler` is something built-in to the Mattermost plugin development framework to run arbitrary code when the browser reconnects to the internet after being disconnected for some amount of time.

I would still like to see the reconnect handler implemented in this PR if you agree it is appropriate @Ashniu123. Thanks for this awesome contribution, I look forward to seeing it in the product!

mickmister:
@Ashniu123 Do you mind also implementing the plugin's [`uninitialize`](https://github.com/mattermost/mattermost-plugin-todo/pull/54) function to include a call to [`unregisterReconnectHandler`](https://developers.mattermost.com/extend/plugins/webapp/reference/#unregisterReconnectHandler)? Other `unregister*` functions should be added there as well. If you are up for it, you can add the other functions as well, or I can create a follow up ticket for the other functions.

Ashniu123:
Ok. I did not know what `registerReconnectHandler` does. Thanks for pointing it out in the docs. 

Just solve one doubt of mine - when the app reconnects to the internet, is the "refresh" websocket event published? Which might suffice. If I do add a `registerReconnectHandler`, which will invoke `refresh` function (right?), then the same function will be executing twice one after the other leading to wastage.

@mickmister

mickmister:
> Just solve one doubt of mine - when the app reconnects to the internet, is the "refresh" websocket event published?

@Ashniu123

No, the `refresh` websocket event is something created by the plugin's server component, and is called when something changes with the user's todos. It has nothing to do with the browser's internet state. I know all of these similar names can be confusing 🙃 

I am confident that all that needs to be done is add a call to `registerReconnectHandler` 🤙 

Here is the function that generates the `refresh` websocket event if you are curious. If you look at the pieces of code that call this function, you will notice that it is always triggered by some event happening with the todos.

https://github.com/mattermost/mattermost-plugin-todo/blob/954f18e3f774bee7e6270a764376fd59a9f59dec/server/plugin.go#L392-L398

DHaussermann:
Hi @Ashniu123  thanks for this PR, it looks great!

I did find an issue where the count of `Incoming Todsos` can become out of date but, I don't think it's an issue with the sidebar not refreshing because the ToDo item is still visible briefly when I open the ToDo list.

Steps:
- Login as test user
- Login as a different user from another client
- As the different user, send a ToDo to the test user
- Observe that the Incoming ToDos has incremented by 1 on the sidebar
- As the different user, Cancel the ToDo item you sent to the test user
**Observed**: 
A. Count does not decrease (waited a while for websocket event in case it's needed) 
B. When I click the incoming icon to open the list the ToDo item is visible for a fraction of a second then disappears and the count is cont is correct.

Can you confirm this issue is unrelated before we merge this PR. I can then create a separate issue for it.
cc @larkox

larkox:
@DHaussermann That seems a problem with the updating indeed. Probably, if you repeat the same steps with the incoming todos RHS opened, you should see the incoming todo not disappearing. Or at least, the count and the todo to be "synchronized" (i.e. if the todo disappear, the number goes down).

If that is the case, it is not related at all with this PR.

DHaussermann:
@larkox Yes, the issue  does not seem to be related to this PR when comparing to master. It seems to be an issue more about how the list of ToDos does not stay current when changes happen from another user and only updated when you open the plugin RHS. The issue was just not very noticeable to the affected user before as the stale item disappears as soon as the list is opened.

I can write up a new issue for this but - It seems very similar to https://github.com/mattermost/mattermost-plugin-todo/issues/34 Please advise.

larkox:
@DHaussermann The issue should be solved in the PR solving #34 , yes. If you take a note to check that also, it would be great.

larkox:
@aaronrothschild Do you mind approving this PR so we can merge it? Thanks!

PR #63, opened 2020-05-07
[GH-58] GolangCI-Lint integration

hectorgabucio:
#### Summary
Integrates GolangCI-Lint.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-todo/issues/58


server/plugin.go
```diff
@@ -162,7 +162,10 @@ func (p *Plugin) handleAdd(w http.ResponseWriter, r *http.Request) {
 
 	receiverMessage := fmt.Sprintf("You have received a new Todo from @%s", senderName)
 	p.sendRefreshEvent(receiver.Id)
-	p.PostBotCustomDM(receiver.Id, receiverMessage, addRequest.Message, issueID)
+	err = p.PostBotCustomDM(receiver.Id, receiverMessage, addRequest.Message, issueID)
```
hanzei, line 165: 0/5 It might be worth to change the signature of `PostBotCustomDM` to not return an `error` and use it as a fire and forget method, because the caller won't block or return on an `error` anyway.

server/command.go
```diff
@@ -178,10 +187,10 @@ func (p *Plugin) runListCommand(args []string, extra *model.CommandArgs) (bool,
 	if len(args) > 0 {
 		switch args[0] {
 		case "my":
```
larkox, line 189: Why not also "my" as a const?

server/list.go
```diff
@@ -318,9 +314,9 @@ func (l *listManager) extendIssueInfo(issue *Issue, ir *IssueRef) *ExtendedIssue
 	case MyListKey:
 		listName = ""
```
larkox, line 315: Here we may want to use a const too.

server/plugin.go
hectorgabucio: You are right, the error was ignored before these changes. I changed the signature

build/custom.mk
```diff
@@ -1 +1 @@
-# Include custom targets and environment variables here
+# Include custome targets and environment variables here
```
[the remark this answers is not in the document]
hectorgabucio: Yep, fixed!

server/command.go
hectorgabucio: Solved!

server/list.go
hectorgabucio: Done

server/bot.go
```diff
@@ -28,16 +28,13 @@ func (p *Plugin) PostBotDM(userID string, message string) error {
...
-func (p *Plugin) PostBotCustomDM(userID string, message string, todo string, issueID string) error {
+func (p *Plugin) PostBotCustomDM(userID string, message string, todo string, issueID string) {
 	channel, appError := p.API.GetDirectChannel(userID, p.BotUserID)
-	if appError != nil {
-		return appError
-	}
-	if channel == nil {
-		return fmt.Errorf("could not get direct channel for bot and user_id=%s", userID)
+	if appError != nil || channel == nil {
+		return
```
hanzei, line 34: Sorry, I was ambiguous with my request. The error should not be ignored but surely logged to allow proper debugging.

server/bot.go
```diff
@@ -28,16 +28,13 @@ func (p *Plugin) PostBotDM(userID string, message string) error {
...
 	}
 
-	_, appError = p.API.CreatePost(&model.Post{
+	_, _ = p.API.CreatePost(&model.Post{
```
hanzei, line 37: Same here

server/command.go
```diff
@@ -213,7 +220,10 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
 	if foreignID != "" {
 		message := fmt.Sprintf("@%s popped a Todo you sent: %s", userName, issue.Message)
 		p.sendRefreshEvent(foreignID)
-		p.PostBotDM(foreignID, message)
+		err = p.PostBotDM(foreignID, message)
```
hanzei, line 223: Would you please also change this method to not return an error similar to `PostBotCustomDM`

server/bot.go
hectorgabucio: Done!

server/bot.go
hectorgabucio: Done!

server/command.go
hectorgabucio: Refactored because it was sharing a lot of logic with PostBotCustomDM

hanzei:
@DHaussermann There are no functional changes in the PR. A general smoke test should be enough.

larkox:
@hectorgabucio Do you mind merging with master and solving the conflicts? Thanks!

hectorgabucio:
> @hectorgabucio Do you mind merging with master and solving the conflicts? Thanks!

done!

hanzei:
Thanks @hectorgabucio for the contribution :tada:

PR #67, opened 2020-05-25
Implement execute_command endpoint

iomodo:
#### Summary
This PR will refactor code and introduce packages like `command`, `api`, `todo` and `plugin` to be able to mock services easily and add unit tests later. 
Also this PR will introduce `execute_command` endpoint to run arbitrary slash commands through the http API


server/command/command.go
```diff
@@ -0,0 +1,110 @@
...
+type logger interface {
+	LogError(msg string, keyValuePairs ...interface{})
+}
+
+type pluginAPI interface {
+	GetUserByUsername(name string) (*model.User, *model.AppError)
+	SendRefreshEvent(userID string)
+}
+
+type todoAPI interface {
```
larkox, line 48: These apis are defined also on service almost the same. Any strong reason to have declared twice?

iomodo: Thats an excellent question. I went for this style, idea is to declare interfaces where they are needed. Package declares an interface which it wants to receive from other package(not other way around). Function's arguments are interfaces and  return types are concrete. This way unit testing is easier. Downside is we might end up with more interfaces in the code. I wanted everyones input on this.

levb: Here's where I arrived to after my own evolution. Basically all "APIs" should be expressed as a struct in, struct+error out. `out` implements a `json.Marshaler` and `md.Markdowner` (+ other "marshaling" interfaces if necessary, for websocket updates and such).

As to the dogma of where interfaces should be declared... I see nothing wrong with a package declaring its own interfaces. If the client wants to use this "standard" definition - it can. Otherwise it can define its own interfaces - all of the signature matching works anyway. 0/5 on this one, but I feel that specifically, `server/todo` should declare its interface for mocking.

server/command/command.go
```diff
@@ -0,0 +1,110 @@
...
+	ChannelID string
+	IssueID   string
+	PostID    string
+}
+
+type logger interface {
+	LogError(msg string, keyValuePairs ...interface{})
+}
+
+type pluginAPI interface {
```
levb, line 43: should we even be introducing it here, or switch to using `mattermost-plugin-api` which is already mockable?

server/command/command.go
```diff
@@ -0,0 +1,110 @@
...
+	GetUserByUsername(name string) (*model.User, *model.AppError)
```
levb, line 44: 3/5 We should not be cascading AppError into our code

iomodo: Yep I missed that

server/command/command.go
larkox: 0/5 but I like interfaces to be declared just once. Whenever there is a change on an interface, it will gracefully propagate to all the places it is needed, having to do a bit less of changes here and there. For example, if I declare a Poster that implements `DM(user, message) error` and then I need to return the post ID, and change it to `DM(user, message) (postID, error)`, I would need to make that changes on all the different interfaces, potentially hiding some errors until later.

Maybe I am thinking too much and there are not real cases like this, but it is something that worry me.

server/command/command.go
iomodo: I didn't want to introduce `plugin-api`, that was not the intent of the PR. But in future we definitely should.

server/command/command.go
iomodo: I'd suggest to introduce exported interfaces when we need them and not before. In general I agree that in future `todo` package will need an interface and mocks as well.

go.mod
```diff
@@ -3,8 +3,9 @@ module github.com/mattermost/mattermost-plugin-todo
 go 1.13
 
 require (
+	github.com/gorilla/mux v1.7.3
 	github.com/mattermost/mattermost-server/v5 v5.20.0
 	github.com/mholt/archiver/v3 v3.3.0
 	github.com/pkg/errors v0.9.1
-	github.com/stretchr/testify v1.5.1
+	github.com/stretchr/testify v1.5.1 // indirect
```
levb, line 10: ?? // indirect

iomodo: I removed `plugin_test.go` that did nothing

hanzei:
@iomodo Heads up that there is a conflict to resolve

server/plugin/plugin.go
```diff
@@ -0,0 +1,86 @@
...
+	"sync"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-plugin-todo/server/api"
+	"github.com/mattermost/mattermost-plugin-todo/server/command"
+	"github.com/mattermost/mattermost-plugin-todo/server/todo"
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/plugin"
+
+	"github.com/pkg/errors"
```
hanzei, line 14: Nit:
```suggestion
	"github.com/pkg/errors"
```

server/plugin/plugin.go
```diff
@@ -0,0 +1,86 @@
...
+	config := p.getConfiguration()
+	if err := config.IsValid(); err != nil {
+		return err
+	}
+
+	botID, err := p.Helpers.EnsureBot(&model.Bot{
+		Username:    "todo",
+		DisplayName: "Todo Bot",
+		Description: "Created by the Todo plugin.",
+	})
```
hanzei, line 44: No bot icon?

server/todo/list.go
```diff
@@ -1,4 +1,4 @@
-package main
+package todo
```
hanzei, line 1: This discussion might be out of scope of the PR:

I've noticed that most plugin have common packages like `command` and `api` which interface with a plugin specific package (in this case `todo`) To unity the naming even more, it could help to also unify the naming of that package e.g. into `app`.

Thoughts @iomodo @levb ?

server/todo/issue.go
```diff
@@ -32,14 +36,14 @@ func newIssue(message string, postID string) *Issue {
 	}
 }
 
-func issuesListToString(issues []*ExtendedIssue) string {
-	if len(issues) == 0 {
+func (issues ExtendedIssues) ToString() string {
```
hanzei, line 39: Naming that `String()` would futil the `Stringer` interface

PR #69, opened 2020-05-27
Add autocomplete

iomodo:
#### Summary
This PR will add slash command autocomplete functionality to a Todo plugin.


server/command.go
```diff
@@ -236,3 +237,34 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
 
 	return false, nil
 }
+
+func getAutocompleteData() *model.AutocompleteData {
+	todo := model.NewAutocompleteData("todo", "[command]", "Avaliable commands: list, add")
```
hanzei, line 242: What about the other commands?

server/command.go
```diff
@@ -236,3 +237,34 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
+
+	add := model.NewAutocompleteData("add", "[message]", "Adds a Todo")
+	add.AddTextArgument("Add a todo", "[message]", "")
```
hanzei, line 245: Does `Add a todo` make clear what kind of message to enter? Maybe even an example is more clear:
```suggestion
	add.AddTextArgument("E.g. be awesome", "[message]", "")
```

server/command.go
```diff
@@ -236,3 +237,34 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
+	todo.AddCommand(add)
+
+	list := model.NewAutocompleteData("list", "[name]", "Lists your Todo issues")
+	items := []model.AutocompleteListItem{{
+		HelpText: "Received Todos",
+		Item:     "in",
+	}, {
+		HelpText: "Sent Todos",
+		Item:     "out",
+	}}
```
hanzei, line 255: From looking a the the options it seams like only `/todo list in` and `/todo list out` are valid commands. But `/todo list` is also one. The current data model doesn't allow `""` as `Item`, but maybe it's helpful to add this to ensure optional arguments can be made clear e.g.:
Screenshot from 2020-05-28 11-55-03

server/command.go
```diff
@@ -236,3 +237,34 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
+
+	pop := model.NewAutocompleteData("pop", "", "Removes the Todo issue at the top of the list")
+	todo.AddCommand(pop)
+
+	send := model.NewAutocompleteData("send", "[user] [todo]", "Sends a Todo to a specified user")
+	send.AddTextArgument("Whom to send", "[@awesomePerson]", "")
+	send.AddTextArgument("Todo message", "[message]", "")
+	todo.AddCommand(send)
+
+	help := model.NewAutocompleteData("help", "", "Display usage")
```
hanzei, line 267: Should the command be in an alphabetical oder?

server/command.go
```diff
@@ -236,3 +237,34 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
+		Item:     "out",
+	}}
+	list.AddStaticListArgument("Lists your Todo issues", false, items)
+	todo.AddCommand(list)
+
+	pop := model.NewAutocompleteData("pop", "", "Removes the Todo issue at the top of the list")
+	todo.AddCommand(pop)
+
+	send := model.NewAutocompleteData("send", "[user] [todo]", "Sends a Todo to a specified user")
+	send.AddTextArgument("Whom to send", "[@awesomePerson]", "")
```
hanzei, line 263: While testing this I've noticed that after picking  a user from the `@` mention picker the autocomplete stays closed. Should I file a JIRA?

server/command.go
iomodo: I talked about this situation with @asaadmahmood, he suggested that user should press escape if they want to end autocomplete earlier, that is why you can not have empty suggestion by design. Probably better UX here is to add `in` and `out` as optional parameters.

server/command.go
iomodo: Technically they can be in any order, I put them in order of the use frequency.

server/command.go
iomodo: This is a known issue and is run by UX as well as @aaronrothschild. If you press space autocomplete will continue suggesting. Fix will need a rework of the at_mention_provider as well as suggestions component.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/mattermost/mattermost-plugin-todo
 go 1.13
 
 require (
-	github.com/mattermost/mattermost-server/v5 v5.20.0
+	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20200527124843-dea705969ca4
```
larkox, line 6: I understand this is targetting certain commit or branch. Couldn't we target a proper version? (5.23? 5.24?)

iomodo: Sure, we will target 5.24 when it is cut.

hanzei: https://github.com/mattermost/mattermost-server/releases/tag/v5.24.0-rc1 can be used

server/command.go
hanzei: It doesn't seam obvious to me that escape should be used. The dialog "forces" me as a user to pick either of these two.

Screenshot from 2020-06-08 13-09-05

@asaadmahmood Thoughts on this discussion?

server/command.go
hanzei: @iomodo Is there a JIRA that we can link here for reference?

server/command.go
iomodo: True, you have to press `esc`

DHaussermann:
Thanks @iomodo for updating this one. 

Since ToDo has a command that takes an optional argument at the end of the command, I did notice a concern with having to type space twice to remove the list. Maybe it's just me but, I find that this may not be obvious to users who want to send without either argument. As this is not specific to any plugin I wrote a Jira task here https://mattermost.atlassian.net/browse/MM-26282 please share any feedback you may have.

server/command.go
hanzei: @DHaussermann Followed up with a ticket https://mattermost.atlassian.net/browse/MM-26282

PR #73, opened 2020-06-05
v0.3.0 Audit fix

larkox:
#### Summary
Apart of running audit fix to remove all vulnerabilities, I have performed some other task trying to improve the general status of the packages. Since there are many changes here, we might consider not including this commit in the v0.3.0 and delay it until 0.4.0.

- Set all package versions as "higher than" (`^` previous to the version number)
- Import the latest version of webapp as a devDependency (used to import a test file) (still has one vulnerability related to jQuery)
- Remove unused dependencies
- Package upgrade
- Set the package to private to remove some warnings
- Set the react version to "detect" to avoid some warnings

There are still some npm warnings around:
- Deprecated packages, mainly imported from other dependencies even in the latest version
- Peer dependencies, which generally conflict with other packages peer dependencies.

#### Ticket Link
None


levb:
I'd say get these into 0.3 and subject to more testing, sooner.

webapp/.eslintrc.json
```diff
@@ -32,7 +32,10 @@
     "beforeEach": true
   },
   "settings": {
-    "import/resolver": "webpack"
+    "import/resolver": "webpack",
+    "react": {
+      "version": "detect"
```
hanzei, line 37: Should this change also go into https://github.com/mattermost/mattermost-plugin-starter-template?

webapp/package.json
```diff
@@ -15,60 +16,46 @@
...
+    "enzyme": "^3.11.0",
+    "enzyme-adapter-react-16": "^1.15.2",
+    "enzyme-to-json": "^3.5.0",
+    "eslint": "^7.1.0",
+    "eslint-import-resolver-webpack": "^0.12.1",
+    "eslint-plugin-import": "^2.20.2",
+    "eslint-plugin-react": "^7.20.0",
+    "jest": "^26.0.1",
+    "jest-canvas-mock": "^2.2.0",
+    "jest-junit": "^10.0.0",
```
hanzei, line 39: Based on the discussion at https://community-daily.mattermost.com/core/pl/hkzq6sunwt8gpr4n14fd6y45bw is seams like exact versions should be used.

webapp/webpack.config.js
```diff
@@ -35,7 +35,9 @@ module.exports = {
                     {
                         loader: 'sass-loader',
                         options: {
-                            includePaths: ['node_modules/compass-mixins/lib', 'sass'],
+                            sassOptions: {
+                                includePaths: ['node_modules/compass-mixins/lib', 'sass'],
+                            },
```
hanzei, line 40: Same question here: Also include this into https://github.com/mattermost/mattermost-plugin-starter-template?

webapp/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.1",
   "description": "This plugin serves as a starting point for writing a Mattermost plugin.",
   "main": "src/index.js",
+  "private": true,
```
hanzei, line 6: Same question here :wink:

webapp/.eslintrc.json
larkox: I would say it should. It mainly remove a warning during the linting.

webapp/webpack.config.js
larkox: This is dependant of the version of one of the libraries (not sure if webpack or babel). If we were to bump the dependencies of the plugin starter template (which I think we should), this will be needed.

webapp/package.json
larkox: This should go to the starter template. It is only to remove some warnings during the building, but make sense to add it.

webapp/webpack.config.js
hanzei: 0/5 maybe just bump the version and include the change?

PR #89, opened 2020-07-02
add plugin info command

svelle:
I wanted to know the version of the plugin on a server I wasn't admin on and realized this plugin have a way to do that.

Added a simple info command to show the installed version.

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -42,6 +42,9 @@ send [user] [message]
 
 	example: /todo send @awesomePerson Don't forget to be awesome
 
+info 
+	Shows plugin information
+
 help
 	Display usage.
 `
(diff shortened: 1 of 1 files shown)


larkox:
@aaronrothschild I just wanted your confirmation on this feature. I see no harm in this, and might be useful. The text should be something like:
`Todo plugin version: 1.4.0`

larkox:
@aaronrothschild Any thoughts on this?

aaronrothschild:
For me, when supporting customers - they may not have access to the system console to see the version and API is kinda funky for some users. 

I'd be more inclined to include the version number on the Help screen rather than adding a new command.

svelle:
@aaronrothschild I thought about that as well, I just added it as an additional command because it's like that in some other plugins as well. I can move it to the help screen no problem.

PR #90, opened 2020-07-03
GH-88 Add settings command to toggle daily reminders

openmohan:
#### Summary
The TODO bot reminds the user with the list of issue daily if it is not empty. With this feature added the reminder preference can be set, so the user can turn `off` or `on` the daily reminder functionality. By default, the value will be set as true.

#### Ticket Link
https://github.com/openmohan/mattermost-plugin-todo/tree/GH-88

####Usage
`/todo settings summary on` or `/todo settings summary off`


server/command.go
```diff
@@ -248,8 +255,33 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
+func (p *Plugin) runSettingsCommand(args []string, extra *model.CommandArgs) (bool, error) {
+	if args[0] == "summary" {
+		var responseMessage string
+		var err error
+
+		if args[1] == "on" {
+			err = p.saveReminderPreference(extra.UserId, true)
+		} else if args[1] == "off" {
+			err = p.saveReminderPreference(extra.UserId, false)
+		}
```
hanzei, line 267: Using a `switch` here would make the code cleaner and would make it easy to add a `default` path.

server/store.go
```diff
@@ -378,3 +384,31 @@ func (p *Plugin) getLastReminderTimeForUser(userID string) (int64, error) {
...
+		return true
+	}
+
+	if preferenceByte == nil {
+		return true
+	}
+
+	preference, err := strconv.ParseBool(string(preferenceByte))
+	if err != nil {
+		return true
```
hanzei, line 410: Nit pick: This error should be logged, just to help with debugging.

server/command.go
```diff
@@ -276,6 +308,9 @@ func getAutocompleteData() *model.AutocompleteData {
 	send.AddTextArgument("Todo message", "[message]", "")
 	todo.AddCommand(send)
 
+	settings := model.NewAutocompleteData("settings", "summary [on] [off]", "Sets the user preference on daily reminder")
```
hanzei, line 311: Maybe out of scope of the PR: It would be great if there are autocomplete options for `on` and `off`

openmohan: Sure, I can cover in this PR.

server/command.go
openmohan: Changed to switch model.

server/store.go
```diff
@@ -378,3 +385,34 @@ func (p *Plugin) getLastReminderTimeForUser(userID string) (int64, error) {
 
 	return reminderAt, nil
 }
+
+func (p *Plugin) saveReminderPreference(userID string, preference bool) error {
+	preferenceString := strconv.FormatBool(preference)
+	appErr := p.API.KVSet(reminderEnabledKey(userID), []byte(preferenceString))
+	if appErr != nil {
+		return errors.New(appErr.Error())
```
hanzei, line 393: Cant this be 
```suggestion
		return appErr
```
instead?

server/store.go
```diff
@@ -378,3 +385,34 @@ func (p *Plugin) getLastReminderTimeForUser(userID string) (int64, error) {
...
+	}
+	return nil
+}
+
+// getReminderPreference - gets user preference on reminder - default value will be true if in case any error
+func (p *Plugin) getReminderPreference(userID string) bool {
+	preferenceByte, appErr := p.API.KVGet(reminderEnabledKey(userID))
+	if appErr != nil {
+		log.Printf("error getting the reminder preference, err=%+v", appErr)
```
hanzei, line 402: Sorry, my previous comment was unclear. The plugin API actually exposes methods for logging. They should be used instead of `log`. You can find an example at https://github.com/openmohan/mattermost-plugin-todo/blob/24e1c44e136db59f6b4044c64ab21bcc36aa88a5/server/plugin.go#L117.

server/command.go
```diff
@@ -248,8 +255,43 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
 	return false, nil
 }
 
+func (p *Plugin) runSettingsCommand(args []string, extra *model.CommandArgs) (bool, error) {
+	if args[0] == "summary" {
```
reflog, line 259: I think there should be a bounds check on 'args' array here

server/command.go
```diff
@@ -248,8 +255,43 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
+		var err error
+		invalid := false
+
+		switch args[1] {
+		case "on":
+			err = p.saveReminderPreference(extra.UserId, true)
+		case "off":
+			err = p.saveReminderPreference(extra.UserId, false)
+		default:
+			invalid = true
```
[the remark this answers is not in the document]
openmohan: yes, we can. :+1: 

Will add the return here.

server/command.go
openmohan: Added length check on arguments

server/store.go
openmohan: simplified as suggested

server/command.go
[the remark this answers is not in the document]
openmohan: removed additional variable `invalid` there and returned in default case.

server/store.go
openmohan: > looks good in general, please add a test

Added test cases for `command.go`->`runSettingsCommand`.

openmohan:
- [ ] Squash commits at end.

DHaussermann:
@hanzei or @aaronrothschild  The 3.0 release is about ready now. I can test this today if you confirm we want it in the release.

hanzei:
It would be quite awesome to get this into the release.

larkox:
@asaadmahmood @aaronrothschild What do you think about along the daily reminder show a message to tell how to stop the daily reminder? Something like:
`If you want to stop receiving daily reminders type /todo settings reminder off`
@DHaussermann This request is a bit last minute, and can be added in a different PR, so if you are already done testing this, I will move this to a different PR.

DHaussermann:
@larkox my personal opinion on this is that since the feature will be included in 3.0, the message should go along with.

Regarding testing progress I'm starting now by figuring out how I can trigger the reminder on demand. But I don't see this addition as invalidating much testing so I am not opposed to including it now.

DHaussermann:
1. Not sure how to test this without actually waiting a full day to test both states of the flag. @larkox I need a way to make the app think I got the ping 25hrs ago. Should I be able to do this my modifying the `reminder_<mmuserid>` key value in the KV store? This did not work for me. I'm not sure this is some kind of timestamp 🙃  Thoughts? 

2. @openmohan there is a test no longer working on this branch.
```
Running golangci-lint
golangci-lint run ./...
server/command_test.go:28:30: `successfull` is a misspelling of `successful` (misspell)
			name:    "Setting summary successfull",
			                          ^
make: *** [golangci-lint] Error 1
```
Can you please see if you can repro and resolve this?

larkox:
@DHaussermann If I understand correctly the underlying logic, you should be able to trigger the daily summary by:
- setting reminder_<mmuserid> to 0
- refreshing the page
- clicking anywhere

If you were not to refresh the page, you may have to wait for 1 hour before the daily reminder get's checked again.

DHaussermann:
Thanks @larkox  I will try this again shortly. 

After team discussion, this will be moved to next release so we can ship this now. Adjusting milestone.

openmohan:
> 1. Not sure how to test this without actually waiting a full day to test both states of the flag. @larkox I need a way to make the app think I got the ping 25hrs ago. Should I be able to do this my modifying the `reminder_<mmuserid>` key value in the KV store? This did not work for me. I'm not sure this is some kind of timestamp 🙃  Thoughts?
> 2. @openmohan there is a test no longer working on this branch.
> 
> ```
> Running golangci-lint
> golangci-lint run ./...
> server/command_test.go:28:30: `successfull` is a misspelling of `successful` (misspell)
> 			name:    "Setting summary successfull",
> 			                          ^
> make: *** [golangci-lint] Error 1
> ```
> 
> Can you please see if you can repro and resolve this?

Sure I will check it out.

PR #105, opened 2020-07-24
Posting response when there are no Todos to pop instead of unkown error

### LINKED ISSUE #92: Pop command throws an "unkwon error" when no todos in the list
Right now, if we have no todos and we run the pop command, we receive the following message:
```
An unknown error occurred. Please talk to your system administrator for help.
```

It should show a more descriptive message like:
```
There are no todos to pop.
```

dantepippi:
This PR fixes issue 92 posting a response message when there are no Todos to pop instead of unkown error.
closes #92

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -216,6 +216,10 @@ func (p *Plugin) runListCommand(args []string, extra *model.CommandArgs) (bool, 
 func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, error) {
 	issue, foreignID, err := p.listManager.PopIssue(extra.UserId)
 	if err != nil {
+		if err.Error() == "cannot find issue" {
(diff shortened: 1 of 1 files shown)


server/command.go
```diff
@@ -216,6 +216,10 @@ func (p *Plugin) runListCommand(args []string, extra *model.CommandArgs) (bool,
 func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, error) {
 	issue, foreignID, err := p.listManager.PopIssue(extra.UserId)
 	if err != nil {
+		if err.Error() == "cannot find issue" {
```
hanzei, line 219: The idiomatic way would be to define an error type e.g. `ItemNotFound` and check for it here using `error.Is`.

hanzei:
@DHaussermann Mind skipping QA review as this is a beta plugin?

PR #110, opened 2020-07-30
Improvement on texts and error behaviours on /todo settings

larkox:
#### Summary
Improve the texts and error messages/behaviours on /todo settings

Current look:

/todo settings
Screenshot from 2020-07-30 10-04-22
/todo settings summ
Screenshot from 2020-07-30 10-05-01
/todo settings summary
Screenshot from 2020-07-30 10-04-35
/todo settings summary on
Screenshot from 2020-07-30 10-05-41
/todo settings summary off
Screenshot from 2020-07-30 10-05-52
/todo settings summary foo
Screenshot from 2020-07-30 10-06-09
/todo settings summary on foo
Screenshot from 2020-07-30 10-06-50

#### Ticket Link
None

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -109,7 +109,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 	isUserError, err := handler(restOfArgs, args)
 	if err != nil {
 		if isUserError {
-			p.postCommandResponse(args, fmt.Sprintf("__Error: %s__\n\nRun `/todo help` for usage instructions.", err.Error()))
+			p.postCommandResponse(args, fmt.Sprintf("__Error: %s.__\n\nRun `/todo help` for usage instructions.", err.Error()))
 		} else {
 			p.API.LogError(err.Error())
 			p.postCommandResponse(args, "An unknown error occurred. Please talk to your system administrator for help.")
@@ -300,7 +300,7 @@ func (p *Plugin) runSettingsCommand(args []string, extra *model.CommandArgs) (bo
 			err = p.saveReminderPreference(extra.UserId, false)
 			responseMessage = "You will stop receiving daily summaries."
 		default:
-			responseMessage = `invalid input, allowed values for "settings summary" are [on] or [off]"`
+			responseMessage = "invalid input, allowed values for \"settings summary\" are `on` or `off`"
 			return true, errors.New(responseMessage)
 		}
 
--- a/server/command.go
+++ b/server/command.go
@@ -278,34 +278,42 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
 }
 
 func (p *Plugin) runSettingsCommand(args []string, extra *model.CommandArgs) (bool, error) {
-	if len(args) != 2 {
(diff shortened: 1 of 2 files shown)


asaadmahmood:
@larkox Can we format the Invalid input on/off like we did above? (in code blocks).

asaadmahmood:
Also, we need a full stop after the end of each error.

larkox:
@asaadmahmood Sure. Here is a screenshot with the on/off corrected, and a few with the full stop at the end of the sentence.

Screenshot from 2020-07-30 10-30-16

Screenshot from 2020-07-30 10-32-33

asaadmahmood:
@larkox 
And can we make that error text use our error color?
And then we can probably even remove the "Error" at the start.

larkox:
@asaadmahmood We are limited by markdown here. If I remember correctly, there is no markdown to change the color of the text (but would be a nice addition! (as long we keep it to a limited set of colors like error, warning, etc...)).

We could go with [Message Attachments](https://docs.mattermost.com/developer/message-attachments.html), and change the color of the bar on the left, but since we the code is in back-end, I am not sure we can access the theme color for error.

server/command.go
```diff
@@ -278,34 +278,40 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
-			p.postCommandResponse(extra, responseMessage)
-			return true, errors.New("invalid argument")
+			responseMessage = "invalid input, allowed values for \"settings summary\" are `on` or `off`"
+			return true, errors.New(responseMessage)
 		}
 
 		if err != nil {
 			responseMessage = "error saving the reminder preference"
-			p.postCommandResponse(extra, responseMessage)
-			return false, err
```
hanzei, line 304: Should this error get logged to help with debugging?

server/command.go
```diff
@@ -278,34 +278,40 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
+			return false, errors.New(responseMessage)
 		}
 
-		responseMessage = fmt.Sprintf("reminder preference changed to %s", args[1])
```
hanzei, line 307: Why did this line got removed?

server/command.go
larkox: Since we are passing it as a non user error, it is logged in line 114.

server/command.go
larkox: We are setting the response message to a more user friendly sentence depending on the selection on lines 298 and 301.

hanzei: Got it :+1:

server/command.go
hanzei: Maybe I'm misunderstand. In L309 a new error only containing `responseMessage` is returned. Should L309 instead wrap the error?

larkox: I see what you mean now. In this case I would log the error here, since the other part will also be shown to the user. What do you think?

hanzei: Sound good :+1:

DHaussermann:
@hanzei or @jfrerich If you have a bit of time, could one of you peer-test this PR using the  peer testing process documented here https://mattermost.atlassian.net/wiki/spaces/INT/pages/619937892/Peer+Testing+Process

PR #119, opened 2020-09-21
Only send commands on command telemetry

larkox:
#### Summary
Right now, when a user types a slash command, we store which command the user is typing. To do so, we register the first word after `/zoom`. If a user types wrongly `/zoom Do groceries` instead of `/zoom add Do groceries` we would send to telemetry `Do`.

This ticket fix this.

#### Ticket Link
None

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -88,7 +88,6 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 		p.trackCommand(args.UserId, "")
 	} else {
 		command := stringArgs[1]
-		p.trackCommand(args.UserId, command)
 		if lengthOfArgs > 2 {
 			restOfArgs = stringArgs[2:]
 		}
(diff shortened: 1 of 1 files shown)


server/command.go
```diff
@@ -104,9 +103,15 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 		case "settings":
 			handler = p.runSettingsCommand
 		default:
+			if command == "help" {
+				p.trackCommand(args.UserId, command)
+			} else {
+				p.trackCommand(args.UserId, "not found")
```
hanzei, line 109: Do we really want to track this? If yes, should we send the `command` as part of the telemetry data?

hanzei (approved):
LGTM

One none blocking question

server/command.go
larkox: I would consider the command part part of the "user data", since may be a mistake like `/todo Go for groceries`, and `Go` is part of the user input data.

I understand that it would be interesting in case there are misspelling or something similar, but I feel safer taking this approach.

hanzei: Good point, makes sense :+1:

levb (approved):

PR #126, opened 2020-10-07
[GH-31] Allow user to edit existing to do items

diode:
#### Summary
Allow user to edit existing to do items

#### Ticket Link
[https://github.com/mattermost/mattermost-plugin-todo/issues/31](https://github.com/mattermost/mattermost-plugin-todo/issues/31)

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -293,47 +293,6 @@ func (p *Plugin) handleUpdate(w http.ResponseWriter, r *http.Request) {
 		p.postReplyIfNeeded(updateRequest.PostID, replyMessage, updateRequest.Message)
 		return
 	}
-
-	if receiver.Id != userID {
-		issueID, err := p.listManager.SendIssue(userID, receiver.Id, updateRequest.Message, updateRequest.PostID)
-
-		if err != nil {
-			p.API.LogError("Unable to send issue err=" + err.Error())
-			p.handleErrorWithCode(w, http.StatusInternalServerError, "Unable to send issue", err)
-			return
-		}
-		p.trackSendIssue(userID, sourceWebapp, updateRequest.PostID != "")
-
-		receiverMessage := fmt.Sprintf("You have received a new Todo from @%s", senderName)
-		p.sendRefreshEvent(receiver.Id)
-		p.PostBotCustomDM(receiver.Id, receiverMessage, updateRequest.Message, issueID)
-
-		replyMessage := fmt.Sprintf("@%s sent @%s a todo attached to this thread", senderName, updateRequest.SendTo)
-		p.postReplyIfNeeded(updateRequest.PostID, replyMessage, updateRequest.Message)
-
-		issue, foreignID, isSender, err := p.listManager.RemoveIssue(userID, updateRequest.PostID)
-		if err != nil {
-			p.API.LogError("Unable to remove issue, err=" + err.Error())
-			p.handleErrorWithCode(w, http.StatusInternalServerError, "Unable to remove issue", err)
-			return
-		}
-		p.trackRemoveIssue(userID)
-
-		userName := p.listManager.GetUserName(userID)
-		replyMessage = fmt.Sprintf("@%s removed a todo attached to this thread", userName)
-		p.postReplyIfNeeded(issue.PostID, replyMessage, issue.Message)
-
-		if foreignID == "" {
-			return
-		}
-
-		message := fmt.Sprintf("@%s removed a Todo you received: %s", userName, issue.Message)
-		if isSender {
-			message = fmt.Sprintf("@%s declined a Todo you sent: %s", userName, issue.Message)
-		}
-
-		p.PostBotDM(foreignID, message)
-	}
 }
 
 func (p *Plugin) postReplyIfNeeded(postID, message, todo string) {
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -295,24 +295,51 @@ func (p *Plugin) handleUpdate(w http.ResponseWriter, r *http.Request) {
 		replyMessage := fmt.Sprintf("@%s attached a todo to this thread", senderName)
 		p.postReplyIfNeeded(updateRequest.PostID, replyMessage, updateRequest.Message)
 		return
+	}else if receiver.Id != userID {
+
+
+		issueID, err := p.listManager.SendIssue(userID, receiver.Id, updateRequest.Message, updateRequest.PostID)
+
+		if err != nil {
+			p.API.LogError("Unable to send issue err=" + err.Error())
+			p.handleErrorWithCode(w, http.StatusInternalServerError, "Unable to send issue", err)
+			return
+		}
+		
+		p.trackSendIssue(userID, sourceWebapp, updateRequest.PostID != "")
+
+		receiverMessage := fmt.Sprintf("You have received a new Todo from @%s", senderName)
+		p.sendRefreshEvent(receiver.Id)
+		p.PostBotCustomDM(receiver.Id, receiverMessage, updateRequest.Message, issueID)
+
+		replyMessage := fmt.Sprintf("@%s sent @%s a todo attached to this thread", senderName, updateRequest.SendTo)
+		p.postReplyIfNeeded(updateRequest.PostID, replyMessage, updateRequest.Message)
+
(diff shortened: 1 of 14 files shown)


webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -213,6 +217,10 @@ export default class SidebarRight extends React.PureComponent {
                                 complete={this.props.actions.complete}
                                 accept={this.props.actions.accept}
                                 bump={this.props.actions.bump}
+                                edit={(issue, e)=>{
+                                    this.props.actions.telemetry('rhs_add', {list: this.state.list});
```
larkox, line 221: Instead of telemetry 'rhs_add' should be edit, right?

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -148,6 +148,10 @@ export default class SidebarRight extends React.PureComponent {
                             complete={this.props.actions.complete}
                             accept={this.props.actions.accept}
                             bump={this.props.actions.bump}
+                            edit={(issue, e)=>{
```
larkox, line 151: Why not follow the same pattern as the other actions? (bump, accept, complete...)

webapp/src/components/root/root.jsx
```diff
@@ -14,6 +14,7 @@ export default class Root extends React.Component {
         visible: PropTypes.bool.isRequired,
         message: PropTypes.string.isRequired,
         postID: PropTypes.string.isRequired,
+        selectedPost: PropTypes.object.isRequired,
```
larkox, line 17: Instead of selected post would be selected issue, right? Since what it is actually selected is a todo issue.

webapp/src/components/root/root.jsx
```diff
@@ -98,7 +108,7 @@ export default class Root extends React.Component {
                             checked={this.state.attachToThread}
                             onChange={this.handleAttachChange}
                         />
-                        <b>{' Add to thread'}</b>
+                        <b>{isEditing ? 'Update thread' : ' Add to thread'}</b>
```
larkox, line 111: If we are updating the todo, do we want to add it to a thread after the update? @aaronrothschild thoughts on this?

larkox: Same will be for sending a todo.

webapp/src/components/root/root.jsx
```diff
@@ -49,9 +54,12 @@ export default class Root extends React.Component {
     }
 
     submit = () => {
-        const {submit, close, postID} = this.props;
+        const {submit, update, close, postID, selectedPost} = this.props;
         const {message, sendTo, attachToThread} = this.state;
-        if (attachToThread) {
+        const isEditing = selectedPost !== null;
+        if(isEditing){
+            update(message, sendTo, postID);
```
larkox, line 61: I feel that when editing a todo, we should not allow to send or attach it to a thread. Nevertheless, I would like to hear the thoughts of @aaronrothschild regarding this.

server/plugin.go
```diff
@@ -225,6 +229,109 @@ func (p *Plugin) handleAdd(w http.ResponseWriter, r *http.Request) {
...
+	if updateRequest.SendTo == "" {
+		_, err = p.listManager.UpdateIssue(userID, updateRequest.Message, updateRequest.PostID)
+		if err != nil {
+			p.API.LogError("Unable to add issue err=" + err.Error())
+			p.handleErrorWithCode(w, http.StatusInternalServerError, "Unable to add issue", err)
+			return
+		}
+		p.trackUpdateIssue(userID, sourceWebapp, updateRequest.PostID != "")
+		replyMessage := fmt.Sprintf("@%s attached a todo to this thread", senderName)
+		p.postReplyIfNeeded(updateRequest.PostID, replyMessage, updateRequest.Message)
```
larkox, line 266: Since I believe we should not be adding the todo to a thread on update, this will not be needed. My question would be: do we want to notify the thread if there are changes in the issue? @aaronrothschild ?

server/plugin.go
```diff
@@ -225,6 +229,109 @@ func (p *Plugin) handleAdd(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}else if receiver.Id != userID {
```
larkox, line 288: The else here is not needed.

server/plugin.go
```diff
@@ -225,6 +229,109 @@ func (p *Plugin) handleAdd(w http.ResponseWriter, r *http.Request) {
...
+
+
+		issueID, err := p.listManager.SendIssue(userID, receiver.Id, updateRequest.Message, updateRequest.PostID)
```
larkox, line 291: Since I believe we should not send to users a todo when updating, this would not be needed. The question is, do we want to modify just my end of the todo, or do we want to modify both ends? (the one in the sender and the one in the receiver) @aaronrothschild ?

larkox (changes requested):
Thank you very much for the contribution! It is looking nice, but there are a few details I would like to iron out.

webapp/src/components/sidebar_right/sidebar_right.jsx
diode: Cannot use the same pattern because on edit two actions are getting called. Instead changed it to a method `edit`

server/plugin.go
levb: 0/5 Too spammy...

levb (approved):
LGTM, but I agree with @larkox comments (no need to post to thread if edited)

webapp/src/components/root/root.jsx
aaronrothschild: > If we are updating the todo, do we want to add it to a thread after the update? @aaronrothschild thoughts on this?

Yes, ideally the the user could provide a status update or reason for the change for example.

webapp/src/components/root/root.jsx
aaronrothschild: > I feel that when editing a todo, we should not allow to send or attach it to a thread. Nevertheless, I would like to hear the thoughts of @aaronrothschild regarding this.

I think we shouldn't send an update automatically.  If the user optionally, provides a message explaining the change then an update should be pushed by default.  

I'm 0/5 on this - as Dani has noted before the user can also easily add a note to the thread as well.

server/plugin.go
aaronrothschild: I agree, probably too spammy.  I suggest an optional update to the thread IF the user provides a text update/reason.  But this is adding some complexity.  

Let's just say we won't update the post if the todo is updated for now.

server/plugin.go
aaronrothschild: I think we should modify BOTH the sent and received text (and indicate it was edited)

aaronrothschild (commented):
I agree with Dani's thoughts on updating the task and not sending notifications whenever it is changed to the thread.

server/plugin.go
diode: Disabled the editing of sent todos.
Not sure how to send a refresh event to the `receiver` to update his `my` list

larkox:
@diode When the PR is ready to be reviewed again, feel free to re-request my review.

diode:
@larkox 
When editing a ToDo assigned to a user, refresh event has to be sent to both the users to update the list. 
After taking updates from master branch I see a change in `sendRefreshEvent` method.
Now I am not sure how to do this for the receiving user to update his `my` list.
So disabled editing for the ToDos assigned to another user. One user can edit his own tasks.

larkox (changes requested):
I left some comments.

Regarding updating the needed lists on the foreign users, it should be simple sending a refresh event to the relevant user and specifying the list to update.

jfrerich:
Thanks for the PR, @diode!  Do you need any clarification from reviewers regarding their feedback?

larkox:
Hi @diode , do you think you will be able to keep working on this PR?

lindy65:
Hi @diode - I'm helping to review stale PRs and, since we haven't heard from you in over 6 months, I'll close this PR for now. 
Please feel free to re-open if you're able to prioritize. Thanks!

I've re-added the "Up for Grabs" label on https://github.com/mattermost/mattermost-plugin-todo/issues/31

//cc @larkox @aaronrothschild

PR #205, opened 2023-03-30
fix rudder key and add missing architectures

trilopin:
#### Summary
After some name changing, MM_RUDDER_WRITE_KEY has become MM_RUDDER_PLUGINS_PROD.

Fix custom.mk so we get the right data.

Additionally, there were missing architectures so I updated the make server command.

#### Ticket Link
no

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ GO_TEST_FLAGS ?= -race
 GO_BUILD_FLAGS ?=
 MM_UTILITIES_DIR ?= ../mattermost-utilities
 DLV_DEBUG_PORT := 2346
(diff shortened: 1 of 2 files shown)


hanzei (changes requested):
https://github.com/mattermost/mattermost-plugin-todo/blob/183fb72f1795b992997cfb583e479596dd408b99/plugin.json#L11-L15 need to get updated for the new architecture

larkox (approved):
Approving not to block the PR. Not sure if I understand everything that is going on on this PR.

hanzei (approved):

PR #207, opened 2023-05-22
Added command for removing any todo

thoratvinod:
#### Summary
Added command for removing any todo

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-todo/issues/195

#### Screenshots
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 dist
 .DS_Store
+node_modules
--- a/server/command.go
+++ b/server/command.go
@@ -125,6 +125,8 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 			handler = p.runListCommand
 		case "pop":
 			handler = p.runPopCommand
+		case "remove":
+			handler = p.runRemoveCommand
 		case "send":
 			handler = p.runSendCommand
 		case "settings":
@@ -327,6 +329,61 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
 	return false, nil
 }
 
+func (p *Plugin) runRemoveCommand(args []string, extra *model.CommandArgs) (bool, error) {
+	message := strings.Join(args, " ")
+	listID := MyListKey
+	issueList, err := p.listManager.GetIssueList(extra.UserId, listID)
+	if err != nil {
+		return false, err
+	}
+
+	var issueID string
+	for _, issue := range issueList {
+		if issue.Message == message {
+			issueID = issue.ID
+		}
+	}
+
+	if issueID == "" {
+		p.postCommandResponse(extra, "Can't find the todo in list")
+		return false, nil
+	}
+
+	issue, foreignID, _, _, err := p.listManager.RemoveIssue(extra.UserId, issueID)
+	if err != nil {
+		return false, err
+	}
+
+	userName := p.listManager.GetUserName(extra.UserId)
+
+	if foreignID != "" {
+		p.sendRefreshEvent(foreignID, []string{OutListKey})
+
+		message := fmt.Sprintf("@%s removed a Todo you sent: %s", userName, issue.Message)
+		p.PostBotDM(foreignID, message)
+	}
+
+	p.sendRefreshEvent(extra.UserId, []string{MyListKey})
+
+	responseMessage := fmt.Sprintf("Removed given Todo (%s).", issue.Message)
+
+	replyMessage := fmt.Sprintf("@%s removed a todo attached to this thread", userName)
+	p.postReplyIfNeeded(issue.PostID, replyMessage, issue.Message)
+
+	issues, err := p.listManager.GetIssueList(extra.UserId, MyListKey)
+	if err != nil {
+		p.API.LogError(err.Error())
+		p.postCommandResponse(extra, responseMessage)
+		return false, nil
+	}
+
+	responseMessage += listHeaderMessage
+	responseMessage += issuesListToString(issues)
+	p.postCommandResponse(extra, responseMessage)
+
+	return false, nil
+}
+
 func (p *Plugin) runSettingsCommand(args []string, extra *model.CommandArgs) (bool, error) {
 	const (
 		on  = "on"
@@ -418,7 +475,7 @@ func (p *Plugin) runSettingsCommand(args []string, extra *model.CommandArgs) (bo
 }
 
 func getAutocompleteData() *model.AutocompleteData {
-	todo := model.NewAutocompleteData("todo", "[command]", "Available commands: list, add, pop, send, settings, help")
+	todo := model.NewAutocompleteData("todo", "[command]", "Available commands: list, add, pop, remove, send, settings, help")
 
 	add := model.NewAutocompleteData("add", "[message]", "Adds a Todo")
 	add.AddTextArgument("E.g. be awesome", "[message]", "")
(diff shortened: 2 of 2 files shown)


mickmister (commented):
@thoratvinod Thanks for this great contribution :+1:

I have one suggestion to de-deplicate/share some existing code handling issue deletion. Please let me know your thoughts on this.

server/command.go
```diff
@@ -327,6 +329,61 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
 	return false, nil
 }
 
+func (p *Plugin) runRemoveCommand(args []string, extra *model.CommandArgs) (bool, error) {
+	message := strings.Join(args, " ")
```
larkox, line 333: If we already do this to create the todo, it is ok, but this may create some issues with todos with extra spaces, if I am not mistaken. For example a todo with a message `two  spaces`, when you try to delete, even if you write `two  spaces`, it will get into this variable as `two spaces`.

larkox, line 333: Having to write down the whole message seems bad UX, but 0/5.

server/command.go
```diff
@@ -440,6 +497,10 @@ func getAutocompleteData() *model.AutocompleteData {
 	pop := model.NewAutocompleteData("pop", "", "Removes the Todo issue at the top of the list")
 	todo.AddCommand(pop)
 
+	remove := model.NewAutocompleteData("remove", "", "Removes the given Todo")
+	remove.AddTextArgument("E.g. be average", "[message]", "")
```
larkox, line 501: "be average"... what a downer XD

larkox (approved):
LGTM, but with a few non blocking concerns.

server/command.go
thoratvinod: True. Maybe we can only take the number of Todo in the remove cmd and delete that specific todo. In that case, I think we should change the list of todos from a bullet list to a numbered list. What do you think about this?

.gitignore
```diff
@@ -1,2 +1,3 @@
 dist
 .DS_Store
+node_modules
```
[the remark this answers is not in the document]
thoratvinod: why this is not needed? If I remove node_modules, I see all node_modules in the git changes
image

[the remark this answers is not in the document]
larkox: All that seems out of place. `node_modules`only should be inside the webapp folder. Maybe @mickmister knows better what is going on here.

server/command.go
larkox: Numbered list is "dangerous". I get my list: `1. todo A, 2. todo B, 3. todo C`. Now I say... delete number 2 (`todo B`). But while I am writing that, another user deletes the first one `todo A`. And I end up deleting the "new second one", `todo C`.

Something I did for the badges plugin was to use the autocomplete to add the badge id (in this case the todo id). May be an option, but 0/5 on how good. I don't like much the way it is done in the badges plugin neither :sweat_smile:

A more elaborate option (and maybe not good neither) is that `/todo delete` opens an interactive dialog with all the todos, and you mark with a checkmark all of the todos you want to delete.

@asaadmahmood Since you were doing some changes on the todo plugin some time ago, not sure if you have any thoughts on the UX for this particular part (slash command for todo deletion).

thoratvinod: @asaadmahmood any suggestion here?

asaadmahmood: @thoratvinod @larkox I think we should have an autocomplete with the todo list.
So when the user types `/todo remove ` an autocomplete opens up, with the todo list (name+description), and then the user can select the one he wants to remove.

When he selects it, the selection can turn into an `id` for simplicity.

thoratvinod: Thanks for the suggestion @asaadmahmood. However, I'm not sure how we can implement this in the code.

After reading the code, I've noticed a few things. Please correct me if I'm wrong:
1. The `OnActivate` method will run only once when the plugin is activated.
2. The `ExecuteCommand` method will run every time a command is entered.
3. The `OnDeactivate` method will run only once when the plugin is deactivated.

We register all plugin commands in the `OnActivate` method, which is executed only once after the plugin is activated. If we want to add autocomplete for the delete command, we would need to fetch the latest todo list, which I believe is not possible.

@mickmister @larkox please let me know if I am missing something or if you have any way to enable autocomplete for delete.

larkox: Take a look to https://github.com/larkox/mattermost-plugin-badges

The idea is to register the autocomplete as something dynamic. Hopefully reading the code in there is enough, but if you need more guidance let me know :smile:

.gitignore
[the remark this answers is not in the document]
thoratvinod: Maybe I mistakenly ran `npm install` or something in the root repository. After removing the node_modules folder from the root directory and running `make deploy` again, there are no node_modules in the root folder. Therefore, I removed the `node_modules` from gitignore.

server/command.go
thoratvinod: lol

server/command.go
thoratvinod: As we are now adopting a different approach, this comment is no longer valid. Therefore, we are resolving this.

server/command.go
```diff
@@ -327,6 +329,61 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
+			issueID = issue.ID
+		}
+	}
+
+	if issueID == "" {
+		p.postCommandResponse(extra, "Can't find the todo in list")
+		return false, nil
+	}
+
+	issue, foreignID, _, _, err := p.listManager.RemoveIssue(extra.UserId, issueID)
```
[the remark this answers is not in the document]
thoratvinod: We would have done that if the `runPopCommand` and `runRemoveCommand` were using the same function of `ListManager`. However, for `runPopCommand`, we are using `ListManager.PopIssue`, which only takes the `userId` parameter. On the other hand, for `runRemoveCommand`, we are using `ListManager.RemoveIssue`, which takes both `userId` and `issueId` as parameters.
If you have another idea or a suggestion on how to integrate these commands more effectively, please let me know.

server/command.go
thoratvinod: Have made changes, also updated the screenshots.

server/command.go
```diff
@@ -330,26 +330,18 @@ func (p *Plugin) runPopCommand(args []string, extra *model.CommandArgs) (bool, e
...
-	if err != nil {
-		return false, err
-	}
-
-	var issueID string
-	for _, issue := range issueList {
-		if issue.Message == message {
-			issueID = issue.ID
-		}
+	if args[0] != "--todo" {
```
mickmister, line 333: We should make sure `len(args) == 2` before accessing `args`

server/plugin.go
```diff
@@ -669,3 +671,40 @@ func (p *Plugin) handleErrorWithCode(w http.ResponseWriter, code int, errTitle s
...
+			Hint:     issue.Message,
+			HelpText: issue.Description,
+		}
+		out = append(out, s)
+	}
+
+	outJSON, err := json.Marshal(out)
+	if err != nil {
+		p.API.LogError("Unable marhsal issues list to json err=" + err.Error())
+		p.handleErrorWithCode(w, http.StatusInternalServerError, "Unable marhsal issues list to json", err)
```
mickmister, line 702:
```suggestion
		p.handleErrorWithCode(w, http.StatusInternalServerError, "Unable marshal issues list to json", err)
```

mickmister (approved):
Just one comment to avoid index out of range issue. Otherwise LGTM :+1:

FYI I'm going on PTO for a few weeks and won't be able to re-review this. Please take this review as my approval other than the index out of range issue.

server/command.go
thoratvinod: Done.

larkox (approved):
LGTM, I will leave @asaadmahmood to consider the UX implications, but that part also looks good to me.

thoratvinod:
@asaadmahmood please have a look.

asaadmahmood (approved):

PR #238, opened 2023-12-03
test(e2e, playwright): added e2e test for /todo help command

rahulsuresh-git:
#### Summary

This introduces a Playwright end-to-end (e2e) test to check if the `/todo help` command functions as expected. We verify the functionality of the various commands available for the Todo plugin.

I've also made a few changes to improve clarity and readability.
Verified that the test passes locally.

image

### DIFF
--- a/e2e/playwright/support/utils.ts
+++ b/e2e/playwright/support/utils.ts
@@ -20,7 +20,7 @@ export const getTeamName = async (client: Client4, userId: string) => {
     return team;
 };
 
-export const getBotDMPageURL = async (client: Client4, userId: string, teamName: string, botUsername: string) => {
+export const getBotDMPageURL = async (teamName: string, botUsername: string) => {
     return `${teamName}/messages/${botUsername}`;
 };
 
--- a/e2e/playwright/support/utils.ts
+++ b/e2e/playwright/support/utils.ts
@@ -1,74 +1,63 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import type {Page} from "@playwright/test";
+import type {Page} from '@playwright/test';
 
-import Client4 from "@mattermost/client/client4";
-import {UserProfile} from "@mattermost/types/users";
+import Client4 from '@mattermost/client/client4';
+import {UserProfile} from '@mattermost/types/users';
+
 
 export const waitForNewMessages = async (page: Page) => {
-  await page.waitForTimeout(1000);
+    await page.waitForTimeout(1000);
 
-  // This should be able to be waited based on locators instead of pure time-based
-  // The following code work "almost" always. Commented for now to have green tests.
-  // await page.locator('#postListContent').getByTestId('NotificationSeparator').getByText('New Messages').waitFor();
+    // This should be able to be waited based on locators instead of pure time-based
+    // The following code work "almost" always. Commented for now to have green tests.
+    // await page.locator('#postListContent').getByTestId('NotificationSeparator').getByText('New Messages').waitFor();
 };
 
-export const getTodoBotDMPageURL = async (
-  client: Client4,
-  teamName: string,
-  userId: string
-) => {
-  let team = teamName;
-  if (team === "") {
-    const teams = await client.getTeamsForUser(userId);
-    team = teams[0].name;
-  }
-  return `${team}/messages/@todo`;
+export const getTodoBotDMPageURL = async (client: Client4, teamName: string, userId: string) => {
+    let team = teamName;
+    if (team === '') {
+        const teams = await client.getTeamsForUser(userId);
+        team = teams[0].name;
+    }
+    return `${team}/messages/@todo`;
 };
 
-export const fillTextField = async (
-  name: string,
-  value: string,
-  page: Page
-) => {
-  await page.getByTestId(`${name}`).fill(value);
+export const fillTextField = async (name: string, value: string, page: Page) => {
+    await page.getByTestId(`${name}`).fill(value);
 };
 
 export const submitDialog = async (page: Page) => {
-  await page.click("#interactiveDialogSubmit");
+    await page.click('#interactiveDialogSubmit');
 };
 
 export const fillMessage = async (message: string, page: Page) => {
-  await fillTextField("post_textbox", message, page);
+    await fillTextField('post_textbox', message, page )
 };
 
 export const postMessage = async (message: string, page: Page) => {
-  await fillMessage(message, page);
-  await page.getByTestId("SendMessageButton").click();
+    await fillMessage(message, page)
+    await page.getByTestId('SendMessageButton').click();
 };
 
-export const cleanUpBotDMs = async (
-  client: Client4,
-  userId: UserProfile["id"],
-  botUsername: string
-) => {
-  const bot = await client.getUserByUsername(botUsername);
+export const cleanUpBotDMs = async (client: Client4, userId: UserProfile['id'], botUsername: string) => {
+    const bot = await client.getUserByUsername(botUsername);
 
-  const userIds = [userId, bot.id];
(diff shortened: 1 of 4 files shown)


mickmister:
@toninis Any ideas why we need to press "Approve and run" on each commit of this forked repo PR? 

Workflows for this repo are here https://github.com/mattermost/mattermost-plugin-todo/tree/master/.github/workflows. Not sure which ones are the "2 workflows awaiting approval"

CleanShot 2023-12-03 at 21 17 44

mickmister:
@toninis Is this rule enforcement new? I know there is a "first-time contributor" check that has been enabled for a while, though I'm not used to seeing this for every commit

e2e/playwright/support/utils.ts
```diff
@@ -1,60 +1,78 @@
...
 export const getPostMessageLocatorId = (postId: string) => {
-    return `#post_${postId} .post-message`;
+  return `#post_${postId} .post-message`;
+};
+
+export const getLastPost = async (page: Page) => {
+  const lastPost = page.getByTestId("postView").last();
+  await lastPost.waitFor();
+  return lastPost;
 };
```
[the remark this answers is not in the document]
rahulsuresh-git: Haven't run into this issue yet, even after multiple runs. I did see some code, probably in the Github plugin, that had this concern mentioned in a comment and had a wait timeout of 1000ms as a temporary fix. I didn't add it here though. Any thoughts @mickmister

e2e/playwright/tests/todo_plugin.spec.ts
```diff
@@ -8,35 +8,71 @@
...
+      await expect(postBody).toContainText('list [listName]');
+
+      // * Assert /todo pop command is visible
+      await expect(postBody).toContainText('pop');
+
+      // * Assert /todo send [user] [message] command is visible
+      await expect(postBody).toContainText('send [user] [message]');
+
+      // * Assert /todo settings summary [on, off] command is visible
+      await expect(postBody).toContainText('settings summary [on, off]');
```
mickmister, line 66: @hanzei Curious what you think about this test. It could be a bit brittle since it's checking the output in this much detail. Essentially if _anything_ changes about the help text (other than the order of commands in the output), this test will break.

We would also want to make this test fail if a new command is added to the help text, so maybe we want to count the number of lines somehow?

I think this is a difficult problem to solve, since we mainly want e2e tests to be "this feature generally works correctly", rather than "this feature's output looks exactly like this"

e2e/playwright/tests/todo_plugin.spec.ts
```diff
@@ -8,35 +8,71 @@
...
+      const c = new pages.ChannelsPage(page);
+
+      // # Run command to trigger help
+      postMessage('/todo help', page);
+
+      // # Grab the last post
+      const post = await getLastPost(page);
+      const postBody = post.locator('.post-message__text-container');
+
+      // * Assert /todo add [message] command is visible
```
mickmister, line 50: I think the comments here are a bit redundant. I think the assertions themselves make it obvious what's happening. Maybe have one "Assert all commands are shown in the help text output" comment

e2e/playwright/tests/todo_plugin.spec.ts
```diff
@@ -8,35 +8,71 @@
...
-
-            // * Assert help is visible
-            await expect(slash.getItemTitleNth(0)).toHaveText('todo [command]');
-
-            await expect(slash.getItemDescNth(0)).toHaveText('Available commands: list, add, pop, send, settings, help');
-            });
-        });
-    },
+  setup: () => {
+    test('checking available commands', async ({ page }) => {
```
mickmister, line 24:
```suggestion
  command_autocomplete: () => {
    test('/todo', async ({ page }) => {
```

[0/5](https://handbook.mattermost.com/company/about-mattermost/list-of-terms) on the exact naming to have here, but `setup` makes me think we are setting something up for other tests, which is not really what's going on here

e2e/playwright/tests/todo_plugin.spec.ts
```diff
@@ -8,35 +8,71 @@
...
+      await expect(slash.container).toBeVisible();
+
+      // * Assert todo [command] is visible
+      await expect(slash.getItemTitleNth(0)).toHaveText('todo [command]');
+
+      await expect(slash.getItemDescNth(0)).toHaveText('Available commands: list, add, pop, send, settings, help');
+    });
+  },
+  commands: () => {
+    test('help', async ({ pages, page, pw }) => {
```
mickmister, line 40: 0/5 on this too

```suggestion
    test('/todo help', async ({ pages, page, pw }) => {
```

e2e/playwright/tests/todo_plugin.spec.ts
```diff
@@ -8,35 +8,71 @@
...
+      const c = new pages.ChannelsPage(page);
+
+      // # Run command to trigger help
+      postMessage('/todo help', page);
+
+      // # Grab the last post
+      const post = await getLastPost(page);
+      const postBody = post.locator('.post-message__text-container');
```
mickmister, line 48: Ideally we don't have selectors like this in the tests themselves

e2e/playwright/tests/todo_plugin.spec.ts
```diff
@@ -8,35 +8,71 @@
...
 import SlashCommandSuggestions from 'support/components/slash_commands';
-import {fillMessage, getTodoBotDMPageURL} from 'support/utils';
+import {fillMessage, getLastPost, getTodoBotDMPageURL, postMessage, } from 'support/utils';
+
+test.beforeEach(async ({ page, pw }) => {
+  const {adminClient, adminUser} = await pw.getAdminClient();
+  if (adminUser === null) {
+    throw new Error('can not get adminUser');
+  }
+  const dmURL = await getTodoBotDMPageURL(adminClient, '', adminUser.id);
```
mickmister, line 18: Can we change this to be:

```suggestion
  const dmURL = await getBotDMPageURL(adminClient, adminUser.id, 'todo');
```

Having any todo plugin specific stuff in the support files will make it more difficult to refactor to shared code. If something is specific to the todo plugin, I think it should be in a file that contains only todo plugin specific things

Note that I removed the `teamName` argument, and added a `botUsername` argument. I think the `teamName` argument was a bit cryptic, since we are usually just passing an empty string it seems. 0/5 on how much value that provides. We do an extra call to the API to get a valid team name when that's not provided, so that's not ideal. Maybe that's something that be calculated ahead of time before any tests are run somehow

e2e/playwright/tests/test.list.ts
```diff
@@ -6,4 +6,8 @@ import core from './todo_plugin.spec';
 
 import '../support/init_test';
 
-test.describe(core.connected);
+// Test if plugin is setup correctly
+test.describe("setup", core.setup);
+
+// Test various plugin actions
+test.describe("actions", core.commands);
```
mickmister, line 13: I think this may read a bit better. It definitely would be nice to be able to pick and choose which commands will be tested from this file

```suggestion
// Test various plugin actions
test.describe('commands', commands.help);
```

e2e/playwright/tests/todo_plugin.spec.ts
rahulsuresh-git: Makes sense. I've reorganized the code, could you review it once? @mickmister

mickmister (commented):
Great work on this @rahulsuresh-git! I have a few comments more comments on the implementation. I think we can make the tests a bit less brittle by making their assertions less specific, and some suggestions on code organization

hanzei:
Removing myself from the list of reviewers until the comments of @mickmister are resolved.

rahulsuresh-git:
Closing this PR as the changes are moved to this single PR: https://github.com/mattermost-community/mattermost-plugin-todo/pull/254.
Note: Comments addressed on the single PR.

PR #271, opened 2025-02-18
Better padding and a visual separator between todo's

mspisars:
Add more padding and a separator between todo's

When looking through a list of more than a couple items of Todo's the list is too crowded.
This is especially true for the `Sent Todo's` list, which does not have a checkbox as a visual indicator that the the next todo is started.
This PR addresses the cramped list with more padding and a bottom border separator.

image


mspisars:
Sent Todos list

image

PR #278, opened 2026-09-23
Update plugin for React 19 in Mattermost v12

nang2049:
## Summary

Mattermost v12 ships React 19.2. This PR makes the Todo webapp use the host's React, ReactDOM, and JSX runtime, and removes the React 19 breakages in reachable UI.

- `plugin.json`: `min_server_version` 6.5.0 -> 12.0.0.
- `webpack.config.js`: externals for `react-dom/client` -> `ReactDOM`, `react/jsx-runtime` -> `ReactJSXRuntime`, `react/jsx-dev-runtime` -> `ReactJSXDevRuntime`. `react` and `react-dom` were already external.
- `react-select` 3.2.0 -> 5.10.2 (and drop `@types/react-select`; v5 ships its own types). v3 calls `findDOMNode`. v4 was rejected because its `LoadingIndicator`, `NoOptionsMessage`, `LoadingMessage`, and `MultiValue` depend on function-component `defaultProps`, which React 19 ignores. v5 has neither problem and lists React 19 as a peer. `autocomplete_selector.tsx` now imports types from the package root. It also passes `isDisabled` instead of the unsupported `disabled` prop; no caller sets that prop.
- `CSSTransition` in `components/modals/modals.jsx` and `widget/todo_toast/todo_toast.tsx` now gets a `nodeRef` attached to the transitioned element.
- `components/buttons/remove.jsx`: function-component `defaultProps` replaced with a default parameter.

These dependencies needed no change: `react-custom-scrollbars` 4.2.1 (webpack uses `lib/`, which has callback refs only, with no string refs or `findDOMNode`) and `react-textarea-autosize` 8.3.3.

## Testing

**Source and bundle inspection**
- Prod and dev `main.js` contain no `jsx-runtime`, `ReactDOM.render(`, `__SECRET_INTERNALS`, or `createFactory`. React and ReactDOM come from the `window` globals.
- The only `findDOMNode` hits are react-transition-group fallbacks that run only when `nodeRef` is missing. The dev build also includes the unused `ReplaceTransition`, pulled in by react-select's barrel import.
- The one `Symbol.for("react.element")` hit is react-is's type constant, used by hoist-non-react-statics. It doesn't create elements.

**Simulated rendering (unit tests)**
- `npm run lint`, `npm run check-types`, and `npm test` pass. The only suite is the manifest test; there are no component tests.

**Real server: Mattermost 12.0.0-rc1 (production build), prod and dev plugin builds**
- Opened and closed the RHS from the app bar. Switched between My and Sent lists; Escape closes the list menu.
- Add Todo form with title and description. In the Assign modal, the async selector loaded `@user1`/`@sysadmin` with avatars, kept `maxLength` 22, and accepted keyboard selection (ArrowDown, Enter). Save works. Escape closes the form.
- Toasts: complete and delete toasts appear with the slide transition. Undo, the close button, and the 5-second auto-dismiss all work.
- Item menu shortcuts `e` (edit; the autosize textarea grows) and `d` (delete).
- Post action "Add Todo" pre-fills the RHS form.
- `/todo add`, `/todo send @user1`, `/todo list out`, and `/todo help` all work.
- sysadmin and user1 accepted, completed, edited, and declined todos sent by each other. Bot `custom_todo` posts render "Won't do" (the new `RemoveButton` default) and "Add to my list" (accept) works.
- Browser console showed no errors from this plugin.

## Known gaps

- `FullScreenModal` (the `Root` component) is unreachable: nothing dispatches `OPEN_ROOT_MODAL` and the reducer has no `rootModalVisible`. It's fixed but untested.
- I didn't see the team sidebar buttons; the test server has one team.
- Menu items swallow Enter because `MenuItem` calls `preventDefault` on every non-Tab key. This predates this PR.

## Build toolchain

A second commit upgrades the webapp build toolchain to match the gitlab and webex plugins. React runtime behaviour and the webpack externals are unchanged.

| Package | From | To |
| --- | --- | --- |
| Node (`.nvmrc`) | 16.20.2 | 24.13.1 |
| webpack / webpack-cli | 4.43.0 / 3.3.12 | 5.108.4 / 5.1.4 |
| babel-loader | 8.1.0 | 9.1.3 |
| @babel/core, preset-env, preset-react, preset-typescript | 7.10.4 | 7.29.7, 7.26.0, 7.26.3, 7.26.0 |
| TypeScript | 3.9.6 | 5.7.3 |
| Jest / babel-jest | 26.1.0 | 29.7.0 (+ jest-environment-jsdom 29.7.0) |
| css-loader / style-loader / sass-loader | 3.6.0 / 1.2.1 / 10.4.1 | 6.11.0 / 3.3.4 / 13.3.3 |
| package-lock.json | lockfileVersion 2 | lockfileVersion 3 (npm 11) |

Other changes:
- The Babel `proposal-*` plugins are replaced with their `transform-*` equivalents. The unused `@babel/polyfill`, `@emotion/core`, and `@emotion/babel-preset-css-prop` are removed.
- webpack config: `devtool: false` for production and `'*'` removed from `resolve.extensions`, both required by webpack 5. `output.clean` is set so stale source maps from earlier builds don't end up in the bundle.
- `npm run debug` now uses `--mode=development` instead of `--mode=none`. With `none`, `process.env.NODE_ENV` stays in the bundle, and webpack 5 no longer polyfills `process`, so the dev bundle would throw in the browser. The debug build keeps its non-eval `source-map` devtool, so it loads under Mattermost's CSP.
- Jest: `testURL` becomes `testEnvironmentOptions.url` and `testEnvironment` is `jsdom`. The `""` entry in `moduleDirectories` is replaced with `modulePaths` (`<rootDir>`, `<rootDir>/src`). Under Jest 29 the empty entry made `require('expect')` resolve to `@jest/expect`, which broke every suite.

Left as is:
- ESLint 7.4, babel-eslint, and @typescript-eslint 3.6 still run on Node 24 and pass with TypeScript 5.7.
- mattermost-redux 5.23 is kept (no major upgrade). It still pulls `redux-offline` from GitHub as a git dependency; the lockfile records it as `git+ssh`, but `npm ci` from a clean `node_modules` passes.
- webpack 5 now warns that `showRHSPlugin` and `closeRootModal` are imported from `actions` but never exported. This predates the upgrade and isn't changed here.

Testing:
- `npm ci`, `npm run lint`, `npm run check-types`, `npm test`, `make dist`, and `MM_DEBUG=1 make dist` all pass on Node 24.11.1.
- On Mattermost 12.0.0-rc1, both the prod and dev builds passed the RHS, async assignee selector, `/todo` commands, complete/delete toasts (undo, close, auto-dismiss), post action, and bot-post accept/"Won't do" flows between sysadmin and user1. The console showed no errors from this plugin.


webapp/package.json
```diff
@@ -68,11 +64,11 @@
     "react-custom-scrollbars": "4.2.1",
     "react-dom": "16.13.1",
     "react-redux": "7.2.0",
-    "react-select": "3.2.0",
```
avasconcelos114: Feels like this is missing the upgrade of React itself plus a few of the other key libs

